Issue #5, opened 2017-07-20
Account creation no longer works


PovilasVytautasC:
Hello,

Recently AcmeClient.NewRegistration stopped working. It gives out this error: 

JsonSerializationException: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.Uri' because the type requires a JSON primitive value (e.g. string, number, boolean, null) to deserialize correctly.

Upon closer investigation, it was noticed that Json in LetEncrypt API (https://acme-staging.api.letsencrypt.org/directory) has changed (link to the change commit: https://github.com/letsencrypt/boulder/commit/8c547473b89c7d86f32a1c6529b3025d753608bb). Not all values are URIs now. There is a new object in the response called "meta".

Currently Certes works with live LetsEncrypt API, but not with staging.

fszlin:
released v1.0.7 to address this issue

PovilasVytautasC:
The library works now. Thank you!

PR #6, opened 2017-07-20
Make Json deserialization more tolerant of changes to API

zak-reynolds:
This is a quick patch for issue #5 -- LE added an object to the response JSON, which couldn't be resolved to a URI. I generalized AcmeDirectory to a Dictionary<string, Object> so that the JSON can be deserialized.


fszlin:
thanks mate

Issue #2, opened 2016-11-04
Add doc for DNS record for dns-01 challenge


fszlin:
Update sample with generating SHA-256 digest of the key authorization for dns-01.

```csharp
    var keyAuthBytes = Encoding.UTF8.GetBytes(keyAuthString);
    var sha256 = new Sha256Digest();
    var hashed = new byte[sha256.GetDigestSize()];

    sha256.BlockUpdate(keyAuthBytes, 0, keyAuthBytes.Length);
    sha256.DoFinal(hashed, 0);

    var dnsValue = JwsConvert.ToBase64String(hashed);
``` 

#1

PR #14, opened 2017-08-16
docs : fixed readme to include "how to" use the DNS Challenge. (Ref issue #2)

guillaumeraymond:
docs : fixed readme to include "how to" use the DNS Challenge.
fix : replaced the "Where().. First()" with a simple call to First() on getting the challenge


guillaumeraymond:
I'm not sure why a change to a .md files failed AppVeyor. Let me know if there is something more I need to do.

fszlin:
It’s fine. It fails because AppVeyor secret variables aren’t available for pull requests. I have added a check to hopefully address this.

Issue #18, opened 2017-09-25
Certes fails to load on .Net Framework v4.7


swiftMessenger:
When attempting to the Certes library in a project utilizing the latest .Net framework, it fails to load (possibly due to conflicting dependency versions related to System.Net.Http and System.ValueTuple).

fszlin:
Fixed in v1.1.3

PR #19, opened 2017-09-25
Added net47 to Certes

swiftMessenger:
Included net47 in Certes.csproj TargetFrameworks as well as in the relevant package reference conditions. I also had to update the System.ValueTuple version to 4.4.0. This should address Issue #18 (Certes fails to load on .Net Framework v4.7).

### DIFF
--- a/src/Certes/Certes.csproj
+++ b/src/Certes/Certes.csproj
@@ -46,7 +46,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.2" />
     <PackageReference Include="System.Net.Http" Version="4.3.2" Condition="'$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net47'" />
-    <PackageReference Include="System.ValueTuple" Version="4.4.0" Condition="'$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net47'" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.1" Condition="'$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'net45'" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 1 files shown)


src/Certes/Certes.csproj
```diff
@@ -45,8 +45,8 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.2" />
-    <PackageReference Include="System.Net.Http" Version="4.3.2" Condition="'$(TargetFramework)' == 'net45'" />
-    <PackageReference Include="System.ValueTuple" Version="4.3.1" Condition="'$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'net45'" />
+    <PackageReference Include="System.Net.Http" Version="4.3.2" Condition="'$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net47'" />
+    <PackageReference Include="System.ValueTuple" Version="4.4.0" Condition="'$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net47'" />
```
fszlin: thanks mate, some questions:
Does net47 need to reference ValueTuple?
I think net47 will fall back to net45 automatically, and it seems the changes making net47 exactly same as net45, could you explain the different?

swiftMessenger:
It looks like net47 does not actually need to reference ValueTuple, so I removed that target framework from the ValueTuple condition and reverted the version number back to the previous value.

The main difference between the net45 and net47 target frameworks, to the best of my understanding, is in how Visual Studio handles nested dependencies. If Certes uses the net45 framework then all of the Certes dependencies have to be built using the net45 framework, which can cause conflicts when the libraries that include Certes also require other net47-only libraries that include the same dependencies as Certes.

Example dependency map:
MyLibrary (net47)
 -> Certes (net45)
 -> -> System.Net.Http (net45)
 -> Library2 (net47)
 -> -> System.Net.Http (net47)

swiftMessenger:
Sweet! Thank you.

PR #21, opened 2017-11-12
Update references in 1.1.2 to work with BouncyCastle 1.8.1.3

bytemaster-0xff:
Version 1.1.2 of Certes nuget does not work with BouncyCastle 1.8.1.3 (at minimum in .NET standard 2.00), get runtime error when calling AddNames.
```
System.MissingFieldException: 
'Field not found: 'Org.BouncyCastle.Asn1.X509.X509Name.DefaultLookup'.'
```
Updating and compiling Certes with updated packages seems to have resolve this issue.

### DIFF
--- a/src/Certes/Certes.csproj
+++ b/src/Certes/Certes.csproj
@@ -43,10 +43,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
-    <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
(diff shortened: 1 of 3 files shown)


bytemaster-0xff:
Hi @fszlin can you take a look at this PR?  It looks like the checks are failing because something expired in the tests.  It looks like we need to update the references to work with the latest version of BouncyCastle.

Thanks

Kevin...

fszlin:
Sorry for the delay, I just fixed the CI error.
I am a bit hesitant about bumping the dependency version, especially when there're breaking changes in the lib...

From the look of it, for netstandard2.0, `X509Name.DefaultLookup` is `IDictionary` in v1.8.1.2, but become `HashTable` in v.1.8.1.3.

fszlin:
could you please try v1.1.3 and see if the issue is resolved?

PR #102, opened 2018-05-04
Fix Certes.Acme.Resource.Challenge error property

### LINKED ISSUE #99: Obsolete property in Challenge resource
When using `IChallengeContext.Resource()` the response is a `Challenge` object with a status. If the status comes back invalid there should be an error object. The current [`IList<object> Errors`](https://github.com/fszlin/certes/blob/master/src/Certes/Acme/Resource/Challenge.cs#L55) in the code is always `null` becouse the property has been replaced since draft 10. It is now a single error with subproblems. It can be fould in [section-8](https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-8)

At the moment I am not able to get the reason why validation vailed.
Is this something you want to fix yourself or should I create a Pull request?

JeffreyVerbeek:
Fixes #99

The ACME response when a challenge validation is invalid has changed in draft 10.
This fix should parse the error correctly.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Export full chain certification in PEM ([#87][i87])
 - Support custom `HttpClient` ([#95][i95])
+- Add `Challenge.Error` for challenge valdidation error ([#99][i99])
 
 ### Changed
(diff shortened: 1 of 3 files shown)


JeffreyVerbeek:
Yes, you are right, didn't think of that. Thanks for fixing that, didn't check github over the weekend.

Issue #173, opened 2018-11-24
EC Account Key encoding issue


webprofusion-chrisc:
In some cases the byte length of an Elliptic Curve point for an ES256/386/512 key can be shorter than the required field length. If this byte array is then base64 encoded it is shorter than the specification expects:
https://tools.ietf.org/html/rfc7518#section-6.2.1.2 and section 2.3.5 of http://www.secg.org/sec1-v2.pdf

This results in the key working OK for account request signing etc (because the encoded points are correct point on the curve) but fails when used as part of auth validations (or anything which uses a hash of the public key) because the base64 encoding is different. See: https://community.letsencrypt.org/t/account-public-key-discrepancy/78233/6

I will follow up with a pull request.

webprofusion-chrisc:
Resolved with PR merge. Closing.

PR #174, opened 2018-11-24
Implement byte padding for Elliptic Curve point encoding as fix for #173

webprofusion-chrisc:
## Description

This change implements byte padding of Elliptic Curve point byte representations to match the required field size.

## Checklist

- [X ] All tests are passing
- [ X] New tests were created to address changes in pr (and tests are passing)
- [X ] Updated README and/or documentation, if necessary

### DIFF
--- a/src/Certes/Crypto/AsymmetricCipherKey.cs
+++ b/src/Certes/Crypto/AsymmetricCipherKey.cs
@@ -10,6 +10,18 @@ namespace Certes.Crypto
 {
     internal class AsymmetricCipherKey : IKey
     {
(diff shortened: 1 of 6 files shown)


webprofusion-chrisc:
Ok, I think this is clean enough to leave as-is, unless @fszlin has any suggestions!

webprofusion-chrisc:
I have added attributes to JsonWebKey and it's deriviatives to enforce the serialization order of properties. One user was getting an incorrect Account Key thumbprint during challenges as the key properties were serializing in a different (non-alphabetical) order resulting in a different json string being hashed.

PR #188, opened 2019-04-29
Made the password optional for PFX builds

roysanchez:
## Description

It's a small change that allows the creation of PFX certificates without a password.

## Checklist

- [x ] All tests are passing
- [x ] New tests were created to address changes in pr (and tests are passing)
- [ ] Updated README and/or documentation, if necessary


webprofusion-chrisc:
Hi, currently you can use an empty string to specify no password, instead of passing null - is this result technically different?

roysanchez:
Hi @webprofusion-chrisc, Yes it's basically the same, just though that by not having to specify the parameter or pass null it would make the intention more clear.

Issue #189, opened 2019-05-07
User-Agent header?


jsha:
Hi! I've been looking through Let's Encrypt logs for clients that aren't sending User-Agent headers. When clients don't send those headers it makes it harder for us to diagnose issues. It looks like certes doesn't currently set a User-Agent. Is that right? If so, would you mind adding one?

Here's [what the RFC says](https://tools.ietf.org/html/rfc8555#section-6.1):

   ACME clients MUST send a User-Agent header field, in accordance with
   [RFC7231].  This header field SHOULD include the name and version of
   the ACME software in addition to the name and version of the
   underlying HTTP client software.

Thanks!

jsha:
This issue is still live, per the linked PR.

jsha:
This issue is still live, per the linked header.

PR #192, opened 2019-06-10
Implement custom UserAgent http header for Certes (#189)

schulz3000:
## Description

Implement custom UserAgent http header for Certes (#189)
UserAgent header has format: Certes\{libversion} (ACME 2.0)

## Checklist

- [x] All tests are passing
- [x] New tests were created to address changes in pr (and tests are passing)
- [x] Updated README and/or documentation, if necessary

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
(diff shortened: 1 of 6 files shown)


webprofusion-chrisc:
Cool! As Certes is a library it would also be useful (essential in some places) to also extend the user agent to include the name/version of the tool/UI etc that's using Certes. Users can override the default http client (skipping this implementation) but it would great if we could just set a string to extend the user agent.

PR #220, opened 2020-05-05
Allow reading PEM's with an AsymmetricCipherKeyPair.

ericwj:
A few minor changes which fix blocking issues using Certes to generate certificates using files from 'Crypt-LE'. This will:

* Allow `KeyAlgorithmProvider.GetKey(string pem)` to work if the `PemReader.ReadObject()` returns an `AsymmetricKeyParameter` instead of a `AsymmetricCipherKeyPair`.
* DELETED: Enable public access to the actual `Pkcs10CertificationRequest` which `CertificationRequestBuilder` uses internally.

### DIFF
--- a/src/Certes/Crypto/KeyAlgorithmProvider.cs
+++ b/src/Certes/Crypto/KeyAlgorithmProvider.cs
@@ -28,8 +28,16 @@ namespace Certes.Crypto
             using (var reader = new StringReader(pem))
             {
                 var pemReader = new PemReader(reader);
-                var pemKey = pemReader.ReadObject() as AsymmetricCipherKeyPair;
-                return ReadKey(pemKey.Private);
+                var untyped = pemReader.ReadObject();
+                switch(untyped)
+                {
+                    case AsymmetricCipherKeyPair keyPair:
+                        return ReadKey(keyPair.Private);
+                    case AsymmetricKeyParameter keyParam:
+                        return ReadKey(keyParam);
+                    default:
+                        throw new NotSupportedException();
+                }
             }
         }
 
--- a/src/Certes/Pkcs/CertificationRequestBuilder.cs
+++ b/src/Certes/Pkcs/CertificationRequestBuilder.cs
@@ -152,7 +152,13 @@ namespace Certes.Pkcs
             };
         }
 
-        private Pkcs10CertificationRequest GeneratePkcs10()
+        /// <summary>
+        /// Generates the CSR.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="Pkcs10CertificationRequest"/>.
+        /// </returns>
+        public Pkcs10CertificationRequest GeneratePkcs10()
         {
             var x509 = new X509Name(attributes.Select(p => p.Id).ToArray(), attributes.Select(p => p.Value).ToArray());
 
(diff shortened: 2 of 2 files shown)


ericwj:
This project will die over that bot closing issues unattended.

ericwj:
This project will still die over that bot closing issues unattended.

ericwj:
Boo bot

webprofusion-chrisc:
Hey, @ericwj  are you using this change in production within an existing tool?

webprofusion-chrisc:
p.s. why Pkcs10CertificationRequest is being exposed as part of this PR? How does certes relate to Crypt-LE?

ericwj:
I shouldn't have mentioned Crypt-LE, sorry about that. Maybe I got an `AsymmetricKeyParameter` because I used some other tool with LetsEncrypt before of which I have files on disk, if Certes isn't able to encounter them normally.

In the latest production code I use, I use the equivalent of `CertificationRequestBuilder.GeneratePkcs10().GetDerEncoded()`.

Anyhow Certes used to be the best thing around in C# before it stopped being maintained and the bot started auto-closing issues. I'm well aware this PR exposes types publicly from BouncyCastle and might not be the preferred way of fixing my issue, but sure let's talk about it.

webprofusion-chrisc:
@ericwj regarding CertificationRequestBuilder.GeneratePkcs10().GetDerEncoded() I think you can call `Generate()` as that will in turn run `GeneratePkcs10()` then return the result of `GetDerEncoded()`. If you can exclude that from this PR and we can determine that exiting tests cover this change then I think we'd be ok to merge it.

This isn't my library but I do use it, and I suspect @fszlin doesn't currently have a lot of time for it (or has no current use for it), which is exactly what's great about open source - projects **can** continue even if the author loses interest. I have one or two PRs of my own to submit shortly.

ericwj:
Sure I am using this PR already from my own fork, but that has all sorts of caveats and annoyances associated with it. Do you have the required permissions such that you are able to accept PR's into this repo and actually publish version updates to the NuGet packages?

Regarding `Pkcs10CertificationRequest`, an older project still uses `GeneratePkcs10` and passes the result to `Org.BouncyCastle.OpenSsl.PemWriter`. I guess it would suffice what this project needs to add an API to export to PEM instead of exposing `GeneratePkcs10`.

webprofusion-chrisc:
@ericwj it would appear that I can merge PRs (I don't remember being given permission) but I can't update nuget. Regarding PemWriter, as mentioned you can get the Der encoded cert via .Generate(), so you should be able to use that to pass to pemwriter?

ericwj:
I don't think so, I'd get a byte array. How would PemWriter know what it represents and how to write that properly?

Plus it defeats not having BouncyCastle stuff in the public API, if I then still need to mess with that myself. `WriteObject` is a total mystery to begin with. It takes `object`, not even `Asn1Encodable`, `System.Object`. It doesn't make sense.

ericwj:
Weird that it fails on MacOS only. Anyway the build SDKs are out of date.

webprofusion-chrisc:
Re-ran and it passed. Good enough for me :)

ericwj:
ty

PR #237, opened 2020-11-16
Handle OrderState "processing"

### LINKED ISSUE #235: Processing of order is ignored
When finalizing an order and the CA takes some time, for whatever reason, it returns `status: "processing"` and the client throws an exception.

I know there is a `Finalize` method to work around. But if the server sends a `Retry-After` the client could retry for `n` times (maybe passed as an optional argument).

>> The order moves to the
>> "processing" state after the client submits a request to the order's
>> finalize
https://tools.ietf.org/html/rfc8555#section-7.1.6

```
    public static async Task<CertificateChain> Generate(
      this IOrderContext context,
      CsrInfo csr,
      IKey key)
    {
      Order order = await context.Resource();
      OrderStatus? status = order.Status;
      OrderStatus orderStatus1 = OrderStatus.Ready;
      if (!(status.GetValueOrDefault() == orderStatus1 & status.HasValue))
      {
        status = order.Status;
        OrderStatus orderStatus2 = OrderStatus.Pending;
        if (!(status.GetValueOrDefault() == orderStatus2 & status.HasValue))
          throw new AcmeException(string.Format(Strings.ErrorInvalidOrderStatusForFinalize, (object) order.Status));
      }
      status = (await context.Finalize(csr, key)).Status;
      OrderStatus orderStatus3 = OrderStatus.Valid;
      if (!(status.GetValueOrDefault() == orderStatus3 & status.HasValue))
        throw new AcmeException(Strings.ErrorFinalizeFailed);
      return await context.Download();
    }
```

### LINKED ISSUE #236: Summary XML documentation for OrderStatus enum
The RFC has a nice list from where the description could lent their values:

```
 o  "invalid": The certificate will not be issued.  Consider this
      order process abandoned.

   o  "pending": The server does not believe that the client has
      fulfilled the requirements.  Check the "authorizations" array for
      entries that are still pending.

   o  "ready": The server agrees that the requirements have been
      fulfilled, and is awaiting finalization.  Submit a finalization
      request.

   o  "processing": The certificate is being issued.  Send a POST-as-GET
      request after the time given in the Retry-After header field of
      the response, if any.

   o  "valid": The server has issued the certificate and provisioned its
      URL to the "certificate" field of the order.  Download the
      certificate.
```

Current state:

image

wiggisser:
## Description

This PR is intended to correctly handle the state `"processing"` of an order, when the the order is to be finalized and downloaded via `IOrderContextExtensions::Generate`. If `orderContext.Finalize()` returns a state of `"processing"`, `Generate` waits for the timespan indicated by the `Retry-After` header (or 1 second if no Retry-After header is found) and then fetches `order.Resource()` again. An optional retryCount can be specified to repeat this wait/fetch multiple times if necessary (default = 1)

Furthermore the documentation of the `OrderStatus` enum is adapted to the descriptions given in the RFC

fixes #235 
fixes #236

## Checklist

- [X] All tests are passing
- [X] New tests were created to address changes in pr (and tests are passing)
- [X] Updated README and/or documentation, if necessary

Thanks for contributing!


webprofusion-chrisc:
Thanks! I'm guessing you might be trying to integrate with ZeroSSL as I had to do something similar a couple of days ago. Which CAs have you tested this against so far?

wiggisser:
@webprofusion-chrisc  No it's not aimed at some particilar CA. I just saw the issue and thought I could have a look into it. And currently I only tested against pebble

PR #239, opened 2020-11-17
Retry on bad nonce

### LINKED ISSUE #238: Retry request on badNonce error
As of [Section 6.5 of RFC 8555](https://tools.ietf.org/html/rfc8555#section-6.5) on receiving a "badNonce" error, a client SHOULD retry the request with a new nonce.

I think I could provide a PR for this in a few days ...

wiggisser:
## Description

This PR is intended to automatically retry a POST request on a badNonce error. 

When a badNonce error is returned, the payload of the request is signed with a the new nonce returned with the response and posted again. An optional retryCount (default = 1) can be specified. 

fixes #238 

## Checklist

- [X] All tests are passing
- [X] New tests were created to address changes in pr (and tests are passing)
- [X] Updated README and/or documentation, if necessary

Thanks for contributing!


webprofusion-chrisc:
Thanks, nice work! This is a fairly extensive change. So the gist of this is that the AcmeContext is now passed into a new version of http Post so that the payload can be re-signed with the new nonce. 

Should the retry preference be set at the AcmeContext level instead? That way you'd set it once on the context and any operation which might hit a bad nonce can use that preference instead of passing it.

wiggisser:
@webprofusion-chrisc Thanks, yes, that's the basic idea ... 

Yes, setting the retry preference on the AcmeContext seems like a good idea. I can change that.

webprofusion-chrisc:
@wiggisser  I think we're both trying to fix the merge at the same time, I think you need `resp = await Context.HttpClient.Post<string>(alternate, payload);` because the next line goes on to use the `resp.Resource`

webprofusion-chrisc:
Or rather, you need:
```
resp =  await Context.HttpClient.Post<string>(Context, alternate, null, false);

```
instead of:
```
 await Context.HttpClient.Post<string>(Context, alternate, null, false);

```

wiggisser:
@webprofusion-chrisc  yes of course. I forgot the assignment thx ...

PR #306, opened 2023-02-16
Add DefaultValueHandling to JsonSerializerSettings

swiftMessenger:
## Description

When the default JsonSerializerSettings are set to _DefaultValueHandling = DefaultValueHandling.Ignore_, certes is unable to generate SSL certificates. The error message is: "**NewOrder request included invalid non-DNS type identifier: type "", value "DomainName"**". The best workarounds I could think of to make certes work in a solution that needs to keep the DefaultValueHandling.Ignore as the default were both lengthly and ugly. However, preventing this from becoming an issue in the first place is a one-line fix.

## New Test

I also started writing a unit test to verify that Certes continues to work despite the default JSON settings, but am not familiar enough with how the tests work in your solution to make much progress. Here is what I had previously if you are interested and willing to add it to the solution:

```
        [Fact]
        public async Task DoesNotRelyOnDefaultJsonSettings()
        {
            //Intentionally setting default settings that might cause problems when used with the AcmeHttpClient
            Newtonsoft.Json.JsonConvert.DefaultSettings = () => new JsonSerializerSettings
            {
                DateFormatHandling = DateFormatHandling.MicrosoftDateFormat,
                DateTimeZoneHandling = DateTimeZoneHandling.RoundtripKind,
                NullValueHandling = NullValueHandling.Ignore,
                DefaultValueHandling = DefaultValueHandling.Ignore,
                Formatting = Formatting.None,
                ObjectCreationHandling = ObjectCreationHandling.Reuse
            };

            //Now confirm that we can place the order despite our bad default settings
            var ctx = new AcmeContext(WellKnownServers.LetsEncryptStagingV2);
            var order = await ctx.NewOrder(new[] { "a.com" });
            var auths = await order.Authorizations();
            Assert.True(auths.GetEnumerator().MoveNext());
        }
```

### DIFF
--- a/src/Certes/Json/JsonUtil.cs
+++ b/src/Certes/Json/JsonUtil.cs
@@ -20,6 +20,7 @@ namespace Certes.Json
                     NamingStrategy = new CamelCaseNamingStrategy()
                 },
                 NullValueHandling = NullValueHandling.Ignore,
+                DefaultValueHandling = DefaultValueHandling.Include,
                 MissingMemberHandling = MissingMemberHandling.Ignore
             };
 


swiftMessenger:
Any update on accepting this pull request? The failed checks were due to environmental factors rather than code and I have so far been unable to find a way to re-trigger them.

This PR is a fix for [Issue #309 ](https://github.com/fszlin/certes/issues/309)

fszlin:
Thanks for the fix. Certes has since moved from Newtonsoft.Json to System.Text.Json with its own serializer options. The file this PR changes no longer exists in that form, and the problem it addressed (#309) is fixed in 4.0.0-beta.1. Closing.

PR #323, opened 2024-02-01
let PfxBuilder support setting the Cert Algorithm of PFX

zhaixiaowai:
## Description

The default Cert Algorithm of Org.BouncyCastle.Pkcs.Pkcs12StoreBuilder is PkcsObjectIdentifiers.PbewithShaAnd40BitRC2Cbc

Android devices do not support RC2's pfx certificate by default, so the modified method here passes the Cert Algorithm into the custom Cert Algorithm of the exported PFX file.

For example: pfxBuilder.Build(friendlyName, pfxPassword, PkcsObjectIdentifiers.PbeWithShaAnd3KeyTripleDesCbc)

## Checklist

- [ ] All tests are passing
- [ ] New tests were created to address changes in pr (and tests are passing)
- [ ] Updated README and/or documentation, if necessary

Thanks for contributing!

### DIFF
--- a/src/Certes/Pkcs/PfxBuilder.cs
+++ b/src/Certes/Pkcs/PfxBuilder.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using System.Linq;
 using Certes.Crypto;
+using Org.BouncyCastle.Asn1;
 using Org.BouncyCastle.Crypto;
 using Org.BouncyCastle.Pkcs;
 using Org.BouncyCastle.Pkix;
--- a/src/Certes/Pkcs/PfxBuilder.cs
+++ b/src/Certes/Pkcs/PfxBuilder.cs
@@ -70,11 +70,14 @@ namespace Certes.Pkcs
         /// </summary>
         /// <param name="friendlyName">The friendly name.</param>
         /// <param name="password">The password.</param>
+        /// <param name="certAlgorithm">The Cert Algorithm</param>
         /// <returns>The PFX data.</returns>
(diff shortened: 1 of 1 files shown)


webprofusion-chrisc:
This is for a PFX you are installing directly into Android or as part a deployed app? The PFX cert algorithm doesn't affect public consumers of the cert if you're just using the cert for a service (e.g. an API server).

zhaixiaowai:
I have embedded Certes into Maui, and I need to issue a certificate on Android, then parse it into X509Certificate2. However, due to Android's lack of support for RC2, I am unable to initialize X509Certificate2 successfully. If pfxBuilder.Build() could support custom Cert Algorithm, then I could parse X509Certificate2 on Android without relying on other API services.

like

`var pfx = pfxBuilder.Build(friendlyName, pfxPassword, PkcsObjectIdentifiers.PbeWithShaAnd3KeyTripleDesCbc) ?? throw new Exception($"pfx create err");
var x509Cert = new X509Certificate2(pfx, pfxPassword) ?? throw new Exception($"pfx restore to cert err");`

fszlin:
Implemented by #371. PFX export now uses AES-256 (PBES2 + PBKDF2/HMAC-SHA256) by default, and keeps the previous 3DES/RC2 mode behind explicit opt-in ( or CLI ) for older consumers.

fszlin:
Correction to the previous note: the explicit opt-in is `PfxBuilder.Encryption = PfxEncryption.Legacy` in code, or `--legacy-encryption` in the CLI.

PR #332, opened 2026-09-20
Add initial GitHub Actions build and package checks

fszlin:
## Summary
- Introduce read-only, SHA-pinned GitHub Actions jobs for PRs and main pushes using SDK 10.0.301.
- Compile the signed library across all targets on Ubuntu 24.04 x64, Windows Server 2025 x64, and macOS 26 ARM64; compile the CLI and both test projects separately with test signing disabled.
- Pack signed library/CLI packages and consume them with an isolated NuGet cache. Exercise RSA/ECDSA key round-trips and CSR generation through the packaged library, then install the CLI and launch --help.
- Provide an explicit manual run_legacy_tests diagnostic that runs the entire net6.0 unit suite with .NET 10 roll-forward and does not suppress failures.
- Update README, AGENTS.md, and CI migration guidance with the exact verification scope and remaining work.
- Restore missing .gitmodules metadata for the existing docs/docstrap gitlink, discovered when the first hosted checkout failed; the referenced upstream commit was verified and checkout now succeeds.

## Verification
Local checks on macOS ARM64 with SDK 10.0.301:
- actionlint .github/workflows/build.yml: passed.
- git diff --check and staged whitespace checks: passed.
- 17 relative documentation links/heading anchors: passed.
- Signed Release library build: all three targets passed, zero warnings/errors.
- CLI and unit/integration test compilation: both test targets passed; existing NETSDK1138 warnings for net6.0 remain.
- Release pack of Certes and dotnet-certes: passed with signing enabled. Existing missing-package-README notices remain.
- Library package consumer: passed strong-name token presence, RSA/ECDSA key round-trips, and CSR generation.
- Packaged CLI install and --help: passed with .NET 10 roll-forward.
- Gitleaks staged-change scan: no leaks found.

Hosted run https://github.com/fszlin/certes/actions/runs/35544493782 at 0d29d32: all three OS builds and package smoke checks passed. The manual-only legacy diagnostic was skipped as designed. A subsequent documentation-only commit records these results.

## Scope and limitations
- A green run means compilation and package consumption, not a passing unit/integration suite or CA interoperability.
- Tests are not automatically executed yet. The recorded legacy baseline is 138 passed / 6 failed because hosted Pebble returns HTTP 401; no assertions or test cases were disabled. Offline fixtures and local Pebble remain follow-up work.
- Production target frameworks, dependency versions, assembly signing configuration, and package identities are unchanged. The new consumer alone targets net10.0.
- No publishing or deployment; required branch checks are not configured yet. Azure Functions helper compilation, net462 runtime execution, and full certificate issuance are outside this initial workflow.

## AI assistance
Implemented and self-reviewed with OpenCode using gpt-6-astra. Checks actually run are listed above. This is not an independent review; human approval is not claimed.


fszlin:
Hosted verification completed for the current head, 9963833: https://github.com/fszlin/certes/actions/runs/35544622135

- Ubuntu 24.04 build: passed.
- Windows Server 2025 build: passed.
- macOS 26 build: passed.
- Legacy unit tests: not run (manual diagnostic only); this is not a passing unit/integration test result.

The initial run exposed missing .gitmodules metadata for docs/docstrap; 0d29d32 restored the verified upstream URL. Both subsequent hosted runs passed all automatic jobs.

AI-assisted verification/self-review: OpenCode using gpt-6-astra. No independent or human review is claimed.

fszlin (commented):
## Review: initial GitHub Actions build and package checks

**Verdict: approve.** The workflow does what the description says, the scope limits are stated accurately rather than oversold, and I reproduced every CI step locally. Comments below are follow-ups, not blockers.

### Verification performed

Reproduced each workflow step locally in a clean worktree at `9963833` (macOS ARM64, SDK 10.0.301):

- `actionlint .github/workflows/build.yml` -> clean.
- `dotnet build src/Certes/Certes.csproj -c Release -p:ContinuousIntegrationBuild=true` -> 0 errors, all three targets, signing enabled.
- `dotnet build test/Certes.Tests.Integration/... -p:SkipSigning=true` -> 0 errors; warnings are NETSDK1138 for `net6.0` plus SourceLink notices (see item 1).
- `dotnet pack` for both `Certes` and `Certes.Cli` at `0.0.0-ci-smoke` -> both packages created; `CERTES_PACKAGE_VERSION` correctly flows through the existing `misc/certes.props` override.
- `dotnet run --project scripts/PackageSmoke` -> "Package consumption passed: assembly identity, RSA/ECDSA key round-trips, and CSR generation."
- `dotnet tool install dotnet-certes --version 0.0.0-ci-smoke` + `--help` -> installed and ran under roll-forward.

Independent checks of specific claims:

- **Action pins resolve to real tags.** `11d5960a...` is exactly `actions/checkout` v4.4.0 and `67a3573c...` is exactly `actions/setup-dotnet` v4.3.1, confirmed against upstream `git ls-remote --tags`.
- **The `.gitmodules` fix is correctly diagnosed.** Run 35544446870 failed on all four jobs with `fatal: No url found for submodule path 'docs/docstrap' in .gitmodules` during checkout's credential-cleanup `git submodule foreach`, which runs even with `submodules: false`. That is the real root cause, and restoring the URL is the minimal fix. I also confirmed the pinned gitlink commit `74b223d` is fetchable from the declared `fszlin/docstrap` URL, so `git submodule update --init` will work; it is not a dangling reference.
- **Hosted runs.** 35544493782 (`0d29d32`) and 35544622135 (head `9963833`) both green across all three OS builds and package smoke, with the legacy diagnostic skipped as designed.

### Design review

The security posture is right for an initial workflow: workflow-level `permissions: contents: read`, `persist-credentials: false`, SHA-pinned actions, `pull_request` rather than `pull_request_target`, per-job timeouts, and no publishing path or secret usage. The `workflow_dispatch` boolean is read through `inputs.run_legacy_tests`, which is the correct context for a typed input.

Keeping the legacy suite manual, with no filters or suppressed assertions, is the honest framing: the six hosted-Pebble failures stay visible instead of being hidden behind an exclusion list. The step summary that explicitly states "build verification only" is a good guard against a green check being misread as a passing test suite.

### Follow-ups

1. **Undisclosed warning class in the package-producing job.** Every build and pack emits `Could not find file 'docs/docstrap/.git'. The source code won't be available via Source Link.` I confirmed these in the hosted log for run 35544622135, in the `Pack signed library and CLI` step, for all three library targets and the CLI. They do not fail the build (MSBuild task warnings are not affected by `TreatWarningsAsErrors`), and main-repository SourceLink is unaffected, but the verification notes list NETSDK1138 and the missing-package-README notices while omitting this one. Either add `submodules: recursive` to the packages job, exclude the gitlink from SourceLink's repository enumeration, or record the warning alongside the others.
2. **CLI compile coverage is implicit and fragile.** The step named "Compile CLI and both test projects" builds only the integration project. That reaches `Certes.Tests` by project reference, and the CLI only through `IncludeCli`, which `Certes.Tests.csproj` sets for `net6.0` alone. The claim is currently true, but if that conditional reference is ever changed the CLI silently stops being compiled while the step name still asserts it. Building `src/Certes.Cli` explicitly would make the guarantee match the label.
3. **The smoke version is duplicated.** `0.0.0-ci-smoke` is hardcoded in `PackageSmoke.csproj` and also set as workflow env and in AGENTS.md. Using `Version="$(CERTES_PACKAGE_VERSION)"` in the `PackageReference` would leave one source of truth; today a change to the env var produces a confusing restore failure.
4. **`cancel-in-progress: true` also applies to pushes to `main`.** Rapid merges will cancel main builds, losing per-commit signal. Since nothing publishes yet the risk is low, but consider restricting cancellation to pull requests.
5. **Pin comments are coarse.** `# v4` next to a v4.4.0 / v4.3.1 SHA makes auditing harder than `# v4.4.0`. Cosmetic.
6. `test/Certes.Func` (`net7.0`) is never compiled by any job. Correctly disclosed as out of scope; worth an explicit line in the replacement plan so it is not forgotten.

### Scope limits I did not verify

Windows and Linux runtime behavior beyond the hosted run's own results, `net462` test execution, the manual legacy diagnostic on hosted infrastructure, and any CA interoperability. Branch protection and required-check configuration are not part of this diff and remain unconfigured, as the PR states.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. This is not an independent human review, and no human approval is claimed.
**Verification:** Local reproduction of all four workflow steps at `9963833` with SDK 10.0.301 on macOS ARM64, plus actionlint, action-SHA-to-tag confirmation, submodule commit reachability, and inspection of hosted run logs 35544446870, 35544493782, and 35544622135.

fszlin:
Addressed the six review follow-ups:
1. Confirmed the SourceLink docs/docstrap/.git warnings in hosted logs and documented them alongside NETSDK1138 and missing-package-README notices. Submodule initialization and SourceLink behavior remain unchanged.
2. Added an explicit CLI build step on every OS.
3. PackageSmoke now consumes CERTES_PACKAGE_VERSION, shared with packing and CLI installation; restore fails with a clear diagnostic when it is missing.
4. Only superseded PR runs are cancelled; push/manual runs use unique concurrency groups to preserve per-commit coverage.
5. Action pin comments now specify v4.4.0 and v4.3.1.
6. Added the Functions helper modernization/retirement decision to the replacement plan.

Local verification: actionlint and whitespace checks passed; explicit CLI compilation and the package consumer passed. Verified missing-version restore fails with the intended error and an alternate version reaches the PackageReference. Hosted checks will rerun on this commit.

AI assistance: Implemented and self-reviewed with OpenCode using gpt-6-astra. No independent or human review is claimed for this follow-up.

fszlin:
Hosted verification for aa0340c passed: https://github.com/fszlin/certes/actions/runs/35545118410

All three OS builds (including the explicit CLI step) and package smoke checks are green. The legacy unit-test diagnostic was not run, as expected for a PR event.

AI-assisted verification: OpenCode using gpt-6-astra; self-review only.

fszlin (commented):
## Re-review of `aa0340c` ("Address build workflow review feedback")

All six follow-ups from my previous review are resolved. Verified each rather than taking the summary at face value.

| # | Item | Status | How I checked |
| --- | --- | --- | --- |
| 1 | SourceLink `docs/docstrap` warnings undisclosed | Fixed | Now listed in `docs/ci-migration.md` with NETSDK1138 and the README notices |
| 2 | CLI compile coverage implicit | Fixed | Explicit `dotnet build src/Certes.Cli/Certes.Cli.csproj -p:SkipSigning=true` step; ran it locally, 0 errors |
| 3 | Smoke version hardcoded | Fixed | `$(CERTES_PACKAGE_VERSION)` plus a guard target; behavior confirmed both ways |
| 4 | `cancel-in-progress` on main pushes | Fixed | Group and cancellation are now PR-scoped |
| 5 | Coarse pin comments | Fixed | All six now read `# v4.4.0` / `# v4.3.1`, matching the SHAs I verified earlier |
| 6 | `Certes.Func` uncovered | Fixed | Added as replacement-plan item 6 with a modernize-or-retire decision |

### Verification

- `actionlint` on the updated workflow: clean.
- Explicit CLI build step reproduced locally: `Build succeeded`, 0 errors.
- **Item 3 checked in both directions.** With `CERTES_PACKAGE_VERSION` unset, restore fails with exactly the intended diagnostic rather than a confusing NuGet resolution error. I then packed and consumed under a deliberately different version, `0.0.0-review-check`, and the consumer passed: the value genuinely flows through to the `PackageReference`, so the previous hardcoding is gone rather than merely relocated. Putting the guard on `BeforeTargets="CollectPackageReferences"` is the right hook, since it fires during restore instead of after a misleading partial failure.
- **Item 4 semantics are correct.** Falling back to `github.run_id` rather than `github.ref` gives push and manual runs a unique group, so cancellation cannot reach them, while PRs keep superseding behavior through the PR number. The conditional `cancel-in-progress` is belt-and-braces on top of that.
- Hosted run 35545118410 at `aa0340c`: all three OS builds and package smoke green, legacy diagnostic skipped as expected for a PR event.

Item 1 is handled the way I'd hoped: the warning is disclosed and attributed to the uninitialized documentation submodule, with resolution deferred to submodule/SourceLink maintenance, instead of being silenced. The added note that these checks do not verify debugger source retrieval is a useful scope limit I had not asked for.

### Trivial

`AGENTS.md` now reads "These checks verify the version supplied by `CERTES_PACKAGE_VERSION`; ... These checks verify local package consumption only", repeating the same opening clause in consecutive sentences. Purely editorial; fold into any later docs touch.

**Verdict: approve.** No outstanding items in this PR's scope.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. This is not an independent human review, and no human approval is claimed.
**Verification:** Reviewed the `9963833..aa0340c` delta; ran actionlint, the explicit CLI build, and the package smoke path under both a missing and an alternate `CERTES_PACKAGE_VERSION` on macOS ARM64 with SDK 10.0.301; inspected hosted run 35545118410. Not verified: Windows/Linux runtime behavior beyond that run's own results, `net462` test execution, the manual legacy diagnostic, and CA interoperability.

PR #333, opened 2026-09-20
Make certificate unit tests offline and run the full suite in CI

fszlin:
## Summary
- Replace hosted certificate acquisition with ephemeral root/intermediate/leaf fixtures generated locally using existing BouncyCastle APIs. Certificates have relative validity dates and matching private keys.
- Preserve all existing test cases and assertions; strengthen PFX theory, CLI JSON/file export, and mocked Azure upload coverage by reopening PFX data and checking keys, certificates, issuer presence, and friendly names. Cover RSA and all three existing ECDSA leaf-key algorithms with RSA-signed CA chains; add a missing-issuer failure test.
- Move IntegrationHelper.cs unchanged to Certes.Tests.Integration and remove the network-backed GetValidCert helper and obsolete shared test collection attributes from unit tests.
- Run all unit tests automatically in every Linux/Windows/macOS build job, without filters or failure suppression. Remove the manual run_legacy_tests opt-in.
- Update development guidance and the baseline to 145 passing unit tests.

## Verification
macOS ARM64, SDK 10.0.301, runtime roll-forward to .NET 10:
- Full net6.0 unit suite: 145 passed, 0 failed, 0 skipped (formerly 138 passed / 6 hosted HTTP 401 failures).
- Both unit and integration projects compile for net6.0 and net462 after moving the helper.
- actionlint and git diff --check: passed.
- Staged Gitleaks scan: no leaks found.
- Confirmed the unit-test source has no references to GetValidCert, IntegrationHelper, GetAcmeUriV2, or the hosted Pebble URL.
- An attempted macOS network sandbox run was inconclusive: it blocked VSTest/testhost communication and aborted before executing tests. A second policy attempt was rejected by sandbox-exec. No OS-level network-isolation success is claimed.

## Scope and limitations
- Production code, dependencies, package configuration, and target frameworks are unchanged.
- Unit tests need no ACME server, containers, or credentials. Restore still needs a NuGet source/cache.
- Hosted ACME integration tests were compiled, not executed; their external services and TLS bypass remain confined to the integration helper.
- net462 runtime behavior and public CA interoperability are not verified. Existing NETSDK1138 and documentation-submodule SourceLink warnings remain.
- Hosted matrix tests and existing package smoke checks will be verified on this PR. Required branch-check configuration remains a follow-up.

## AI assistance
Implemented and self-reviewed with OpenCode using gpt-6-astra. AI assisted with implementation and verification; no independent review or human approval is claimed.


fszlin:
Hosted verification passed for 09490e1: https://github.com/fszlin/certes/actions/runs/35545870442

- Ubuntu 24.04, Windows Server 2025, and macOS 26: all builds and full unit-test runs passed.
- Each OS: 145 passed, 0 failed, 0 skipped, with .NET 10 roll-forward.

The six former hosted-service failures are now covered by local fixtures; no test filters or failure suppression are used. Hosted integration tests and net462 runtime tests remain unexecuted.

AI-assisted verification/self-review: OpenCode using gpt-6-astra; no independent or human review is claimed.

fszlin (commented):
## Review: offline certificate fixtures and full unit suite in CI

**Verdict: approve.** This removes the last hosted-service dependency from the unit suite and, contrary to what "replacing real certificates with synthetic ones" usually implies, it *increases* coverage substantially. Details and measurements below, followed by gaps worth recording.

### Coverage impact: measured, not estimated

The review request asked specifically whether the test change affected coverage, so I measured it rather than reasoning about it. I ran the `net6.0` unit suite with `--collect:"XPlat Code Coverage"` on `origin/main` (`bfcc845`) and on this branch (`09490e1`), same machine, SDK 10.0.301, roll-forward to .NET 10, and diffed the Cobertura reports.

| | main | this PR |
| --- | --- | --- |
| Tests | 138 passed / 6 failed | **145 passed / 0 failed** |
| Line coverage | 81.86% | **88.29%** |
| Branch coverage | 72.85% | **79.76%** |

**No class lost coverage.** Every class whose numbers changed went up:

| Class | main | this PR |
| --- | --- | --- |
| `Certes.Pkcs.PfxBuilder` | 27/68 | 65/68 |
| `Certes.Cli.Commands.AzureAppCommand` | 32/72 | 70/72 |
| `AzureAppCommand/<UploadCertificate>` | 0/16 | 16/16 |
| `AzureAppCommand/<FindCertificate>` | 0/15 | 15/15 |
| `Certes.Cli.Commands.CertificatePfxCommand` | 34/71 | 69/71 |
| `Certes.CertificateChainExtensions` | 23/34 | 34/34 |

The reason for the jump is worth stating plainly: on `main` those six tests threw at `GetValidCert()` before reaching the code under test, so `PfxBuilder.Build` and the two CLI command bodies were almost entirely *uncovered* while still appearing in the suite. The old coverage was partly illusory. Two async state machines went from 0 covered lines to fully covered.

The old tests were also weaker than they looked in a second way: `CanCreatePfxChain` and `CanCreatePfxWithoutChain` had **no assertions at all** — they built a PFX and discarded it — and they constructed the builder with `KeyFactory.NewKey(alog)`, a freshly generated key unrelated to the certificate. So the pre-existing tests never verified key/certificate correspondence. The fixture's matching key plus `AssertPfx` (private key DER equality, leaf DER equality, chain[0] identity, intermediate presence) is a real strengthening, not a like-for-like swap.

### Verification performed

- Full suite on this branch: **145 passed, 0 failed, 0 skipped**, reproducing the stated result.
- Same suite on `main` for comparison: 138 passed, 6 failed on hosted HTTP 401.
- Integration project still compiles after `IntegrationHelper.cs` moved: 0 errors.
- Reproduced CI's exact step coupling (`dotnet build` the integration project, then `dotnet test --no-build --no-restore`): passes, so the `--no-build` optimization is sound against the preceding step's outputs.
- Confirmed no residual network surface in `test/Certes.Tests`: no `GetValidCert`, `IntegrationHelper`, `GetAcmeUriV2`, or hosted host names; remaining HTTP references are `Mock<IAcmeHttpClient>` and literal URIs used as data.
- Confirmed `AddIssuers(byte[])` did **not** lose coverage despite its direct call being deleted from `PfxBuilderTests` — the CLI test's `--issuer ./issuers.pem` path still reaches `CertificatePfxCommand.cs:81`. Similarly `./Data/leaf-cert.pem` is still consumed by `CertificateChainTests` and `CertificatePemCommandTests`, so it is not orphaned.
- Hosted run 35545870442 at `09490e1`: all three OS builds, now including the unit-test step, plus package smoke, green.

### Gaps worth recording (not blockers)

1. **The fixture chain is RSA-signed regardless of the key algorithm under test.** Root and intermediate keys are always RS256 and `Issue` always uses `SHA256WITHRSA`. The ES256/384/512 theory cases therefore exercise ECDSA only as the PFX *subject* key, never an ECDSA-signed certificate. That matches the old hosted chain, so it is not a regression, but "Cover RSA and all three existing ECDSA algorithms" in the description reads broader than what is actually verified. Worth one sentence of scope limiting.
2. **Chain-shape realism is reduced.** The fixture is a single clean, correctly ordered root/intermediate/leaf. Alternate and cross-signed chains, out-of-order or extraneous issuer bundles, and duplicate subject DNs are not exercised — exactly the area `AGENTS.md` flags as a `CertificationStore` pitfall (issuers indexed by subject DN). Previously nothing covered this either, and local Pebble is already the stated follow-up; it just should not be assumed covered because the suite is now green.
3. **The embedded root resources are now unit-untested.** `dst-root-ca-x3.pem`, `fake-le-root-x1.pem`, and `isrg-root-x1.pem` ship in `src/Certes`, and `AddTestCerts()` now exists only in the integration project, which CI does not run. Given that `AGENTS.md` lists the stale embedded roots as a known defect, having zero automatic coverage of that path is worth an explicit note.
4. **Minor, fixture hygiene.** `AssertPfx` calls `Leaf.Verify(Intermediate.GetPublicKey())` and `Intermediate.Verify(Root.GetPublicKey())`, which validate the *fixture's* own construction rather than any Certes output. They belong in the constructor; in the assertion helper they slightly overstate what the product is being held to.
5. **Minor, CI coupling.** The test step relies on the preceding build step having used matching properties, since it passes `--no-build --no-restore`. It works today (I verified the exact sequence), but a future change to `SkipSigning` on only one of the two steps would fail in a confusing way. A comment would help.

Removing the `run_legacy_tests` input and the `legacy-tests` job is the right cleanup now that there is nothing to quarantine, and running the suite on all three runners without filters is the correct default. The documentation updates are consistent with the new state, including moving the `IntegrationHelper.cs` path reference and the note that "offline" refers to test execution rather than NuGet restore.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. This is not an independent human review, and no human approval is claimed.
**Verification:** Full unit suite on both `main` and `09490e1` with Cobertura coverage collection and a per-class diff; integration-project compilation; reproduction of the CI build-then-`--no-build`-test sequence; residual-network grep of the unit-test tree; inspection of hosted run 35545870442. Not verified: `net462` runtime execution, hosted ACME integration tests, OS-level network isolation (the PR correctly declines to claim this too), and CA interoperability.

fszlin:
Addressed the review follow-ups:
- Moved fixture signature verification into construction, keeping AssertPfx focused on exported output.
- Documented RSA-signed fixture chains, RSA/ECDSA leaf-key scope, and remaining complex-chain coverage gaps.
- Clarified that embedded-root fallback is still exercised by CertificateChainTests.CanGenerateFullChainPemWithKey (DST Root CA X3 PEM export). The new fixture does not cover that fallback, and coverage is not comprehensive across embedded roots/PFX paths.
- Added a workflow comment about keeping configuration, target, and SkipSigning aligned when reusing build outputs.
- Recorded the successful three-OS unit-test run in the migration notes.

Verification: all 12 affected/related tests passed locally with .NET 10 roll-forward; actionlint and whitespace checks passed. Hosted checks will rerun.

AI assistance: Implemented and self-reviewed with OpenCode using gpt-6-astra. No independent or human review is claimed for these follow-ups.

fszlin:
Hosted checks passed for f75c69c: https://github.com/fszlin/certes/actions/runs/35546343595 — all three OS build/unit-test jobs and package smoke checks are green.

fszlin (commented):
## Re-review of `f75c69c` ("Clarify fixture coverage and address test review feedback")

All five follow-ups are resolved. More importantly, **one of my previous points was wrong, and this commit is right to push back on it.**

### Correction to my previous review

I claimed the embedded root resources now have "zero automatic coverage" because `AddTestCerts()` moved to the un-run integration project. That was overstated, and the response correctly narrows it to the fixture rather than the suite. I verified the counter-claim instead of accepting it:

- `CertificateChainTests.CanGenerateFullChainPemWithKey` feeds `./Data/cert.pem`, which contains **2** certificates, and asserts an output of three ending in DST Root CA X3. That root is not in the input, so it can only come from `CertificateStore.embeddedCertificates`, which enumerates `.pem` manifest resources from the `Certes` assembly.
- Measured directly: `Certes.Pkcs.CertificateStore` (in `CertificationStore.cs`) is **43/43 lines covered on this branch — and 43/43 on `main` as well**. So embedded-root resolution was covered before, is covered now, and this PR does not change it either way.

My underlying concern was about `AddTestCerts()` and the *fake LE* root specifically, but I generalized it into a claim about the whole embedded-root path that the measurement does not support. The new `docs/ci-migration.md` wording is the accurate version: fallback is exercised for PEM export, but it is not comprehensive across every embedded root or the PFX path.

### Remaining follow-ups

| # | Item | Status |
| --- | --- | --- |
| 1 | RSA-signed chains presented as broader ECDSA coverage | Fixed — scope stated explicitly |
| 2 | Chain-shape realism gaps | Fixed — cross-signing, unordered/extraneous issuers, duplicate DNs named as uncovered |
| 3 | Embedded roots | Corrected by author; my claim was wrong (above) |
| 4 | Fixture self-verification inside `AssertPfx` | Fixed — moved into the constructor |
| 5 | CI `--no-build` coupling | Fixed — comment added |

Item 4 is done the right way: `Leaf.Verify(...)` / `Intermediate.Verify(...)` now run once at construction, so a malformed fixture fails fast, and `AssertPfx` asserts only against exported output. That is a cleaner separation than what I suggested, since it also covers fixtures whose PFX is never built.

### Verification

- Full suite at `f75c69c`: **145 passed, 0 failed, 0 skipped**. Moving the verification calls into the constructor did not change results.
- Overall coverage unchanged from my previous measurement: line **88.29%**, branch **79.76%** (vs 81.86% / 72.85% on `main`). `PfxBuilder` remains 65/68.
- `CertificateStore` 43/43 on both `main` and this branch, as above.
- Hosted run 35546343595 at `f75c69c`: all three OS build/unit-test jobs and package smoke green.
- The migration note now cites the completed three-OS run rather than "hosted verification pending", which matches run 35545870442.

**Verdict: approve.** No outstanding items. Net effect of the PR: the unit suite loses its hosted-service dependency, gains 7 cases, goes from 6 failures to none, and raises line coverage by 6.4 points and branch coverage by 6.9, with no class regressing.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. This is not an independent human review, and no human approval is claimed.
**Verification:** Re-ran the full unit suite and Cobertura coverage collection at `f75c69c`, plus a targeted `CertificateStore` coverage comparison against `main` to check my own earlier claim; inspected `CanGenerateFullChainPemWithKey` inputs and `CertificateStore`'s embedded-resource loading; reviewed hosted run 35546343595. Not verified: `net462` runtime execution, hosted ACME integration tests, and CA interoperability.

PR #334, opened 2026-09-21
Move development, CLI, and modern tests to .NET 10

fszlin:
## Summary
- Require a .NET 10 SDK for local development without global.json; Actions installs the latest 10.0.x SDK and records dotnet --info (plus 8.0.x for native .NET 8 package smoke).
- Replace the modern net6.0 library target with net10.0 and net8.0, retaining netstandard2.0 and net462. Move CLI and modern unit/integration test targets to net10.0, retaining net462 test compilation and CLI test inclusion.
- Remove runtime roll-forward from CI and documented development commands.
- Preserve exception serialization members but mark them obsolete on modern .NET to match SYSLIB0051; replace deprecated CLI certificate loading with X509CertificateLoader and dispose loaded certificates.
- Run package consumers natively against net10.0 and net8.0 assets, plus compile a net6.0 consumer through the netstandard2.0 compatibility asset. MSBuild asserts the selected asset and package version on all three targets. Keep the CLI installation smoke test native on .NET 10.
- Update usage, changelog, agent guidance, and migration status.

## Compatibility decisions
- The next CLI package requires .NET 10. This is an intentional runtime requirement change.
- .NET 8/9 consumers select net8.0; .NET 6/7 consumers select netstandard2.0; no dedicated net6.0 asset remains. The retained net462 asset and assembly/package identity are unchanged.
- Exception formatter-serialization APIs remain present; the net8.0 and net10.0 assets carry modern obsolescence annotations. No blanket warning suppression or BinaryFormatter re-enablement.
- Dependency versions are unchanged. The net7.0 Functions helper remains outside CI pending local Pebble replacement.

## Verification
Local macOS ARM64, SDK 10.0.301 / .NET 10.0.9:
- Native net10.0 unit suite: 145 passed, 0 failed, 0 skipped; no DOTNET_ROLL_FORWARD.
- Both test projects compile for net10.0 and net462.
- Signed Release pack of library (all three targets) and CLI passed with an empty temporary NuGet cache and version 0.0.0-net10-smoke.
- Packaged .NET 10 consumer passed asset selection, strong-name token presence, RSA/ECDSA key round-trips, and CSR generation.
- net6.0 consumer compilation passed; project.assets.json confirmed lib/netstandard2.0/Certes.dll selection (net10.0 selected lib/net10.0/Certes.dll).
- Packaged CLI installed and --help ran directly on .NET 10.
- Current NuGet vulnerability audits reported no vulnerable packages in either shipping project; CLI deprecation audit still reports legacy Azure/identity dependencies. This is a target-graph observation, not a dependency upgrade or proof of no security issues.
- actionlint, whitespace checks, and staged secret scan passed.

Known limitations: SourceLink warnings for the uninitialized documentation submodule and package README notices remain. No net462 runtime tests, net6.0 consumer execution, live Azure interoperability, or CA integration verification. Hosted run 35547256549 passed all four checks at 6b3b7d0. Checks are rerunning after the .NET 8 and 10.0.x follow-up; see review replies for expanded local verification.

## Repository setting applied separately
After verifying successful main run 35546812954, required the three Build OS checks and Package smoke checks on main, restricted to the GitHub Actions app, with up-to-date branches required. Existing administrator bypass and other branch protections were preserved. These settings are effective already and are not applied by merging this diff.

## AI assistance
Implemented and self-reviewed with OpenCode using gpt-6-astra. AI assisted with implementation, verification, and required-check configuration. No independent or human review is claimed.


fszlin:
Hosted verification passed for 6b3b7d0: https://github.com/fszlin/certes/actions/runs/35547256549

- Ubuntu 24.04, Windows Server 2025, and macOS 26: all builds and native net10.0 unit tests passed (145 passed, 0 failed, 0 skipped per OS).
- Package smoke checks passed: signed packages, native .NET 10 consumer/CLI, and compile-only .NET 6 compatibility consumer.
- All four checks are now required on main.

AI-assisted verification: OpenCode using gpt-6-astra (self-review). No net462 runtime or CA interoperability verification is claimed.

fszlin:
Removed global.json as requested. Local SDK selection follows the installed .NET environment; both Actions jobs explicitly install SDK 10.0.301. Updated README, AGENTS.md, changelog, and migration guidance. Solution-format migration remains a separate follow-up.

Verification: actionlint and whitespace checks passed; local dotnet --version resolves to 10.0.301. Hosted checks are rerunning.

AI assistance: OpenCode using gpt-6-astra, implementation and self-review.

fszlin (commented):
## Review: move development, CLI, and modern tests to .NET 10

**Verdict: approve, with one packaging decision worth reconsidering before release** (a `net8.0` asset, discussed below). This is the first PR in the revival to touch production code, so I reviewed the source changes line by line and reproduced the packaging behavior rather than relying on the description.

### Production code changes

All four are correct as written:

- **`X509CertificateLoader.LoadCertificate` in `AzureAppCommand`.** Input is always `cert.Certificate.ToDer()`, a single DER certificate, so this is behaviorally equivalent to `new X509Certificate2(byte[])` and strictly *safer*: the loader refuses PKCS#12/PFX content instead of silently accepting it. Adding `using` also fixes a genuine pre-existing `X509Certificate2` leak.
- **No use-after-dispose.** I checked both sites. In the command, `Thumbprint` is captured into a `string` immediately and `x509Cert` is never touched again. In `AzureAppCommandTests`, `cert.Thumbprint` is read eagerly inside the `Setup(...)` object initializer and again in an assertion, both within the `using var` scope.
- **Conditional-compilation guards are consistent with each file's TFM set.** `src/Certes` and the CLI test need no guard for the loader (`net10.0`-only compilation; `Cli\**` is excluded from the `net462` test build via the existing `IncludeCli` item group), while `AcmeContextIntegration.cs` compiles for `net462` too and is correctly guarded with `#if NET9_0_OR_GREATER`.
- **SYSLIB0051 annotations build cleanly under `TreatWarningsAsErrors`** because each obsolete member only calls other obsolete members (`AcmeRequestException`'s ctor chains to `AcmeException`'s obsolete ctor; `GetObjectData` calls the already-obsolete base). Marking rather than deleting is the right call — the members stay available on `netstandard2.0`/`net462`.

### Verification performed

Clean worktree at `80ce393`, macOS ARM64, SDK 10.0.301:

- Signed Release library build, all three targets: 0 errors, 0 warnings.
- Unit suite **natively on `net10.0`, no roll-forward: 145 passed, 0 failed, 0 skipped**.
- `net462` test compilation: 0 errors.
- Signed Release pack of library and CLI; packaged `net10.0` consumer passed; packaged CLI installed and `--help` ran **without** `DOTNET_ROLL_FORWARD`.
- **Asset selection independently confirmed** by parsing `project.assets.json` per target rather than trusting the compile step: `net10.0 -> lib/net10.0/Certes.dll`, `net6.0 -> lib/netstandard2.0/Certes.dll`. The compatibility claim holds.
- **Dependency groups compared in the generated nuspec:** `net10.0` and `.NETStandard2.0` are identical (`Newtonsoft.Json 13.0.2`, `Portable.BouncyCastle 1.9.0`). Only `net462` adds `System.ValueTuple`. So consumers falling back to `netstandard2.0` get the same dependency closure.
- **Collateral check on the one untested project:** `test/Certes.Func` (`net7.0`) still builds against the library through the `netstandard2.0` asset after `net6.0` was dropped. No hidden break.
- Required branch protection confirmed via the API: `Build (ubuntu-24.04)`, `Build (windows-2025)`, `Build (macos-26)`, `Package smoke checks`, strict mode on, restricted to the Actions app. This PR changes step names but **not** job names, so the required contexts still resolve — worth stating explicitly, since a job rename here would have silently blocked all future merges.
- Hosted run 35548024969 at `80ce393`: all four required checks green.

### The one decision I'd revisit: no `net8.0` asset

Dropping `net6.0` is clearly right. Replacing it with `net10.0` *only* means every consumer on .NET 8 — currently LTS and supported until November 2026, and probably the single most common consumer TFM today — now binds to `netstandard2.0`.

I checked how much this actually costs, and the answer is "almost nothing": identical dependency groups, and the only source-level difference between the assets is the three `[Obsolete]` attributes, so .NET 8 consumers merely lose SYSLIB0051 annotations they cannot act on anyway. So this is not a correctness problem and I am not blocking on it.

It is still worth a deliberate decision rather than a side effect, because `netstandard2.0` is the lowest-fidelity binding a modern consumer can get, and adding `net8.0;net10.0` costs one token in `TargetFrameworks`. If the intent is "modern consumers should get a modern asset", `net8.0` belongs in the list until it goes out of support.

### Other follow-ups

1. **The CLI runtime jump from .NET 6 to .NET 10 is the real user-visible break in this PR** — larger than the library change, since `dotnet tool install dotnet-certes` will simply fail for anyone on .NET 8. The changelog entry covers it, but this should drive the version decision at release time (`AssemblyVersion` is still `3.0.0`), and the release notes should lead with it rather than leaving it in a "Changed" bullet.
2. **The `net6.0` probe asserts nothing.** The `net10.0` consumer verifies its asset via `TargetFrameworkAttribute`, but the `net6.0` leg is compile-only, so CI would not catch a regression that changed which asset a down-level consumer resolves. I verified it by hand this time; encoding it as an MSBuild assertion on `@(ReferencePath)` would make CI enforce what the PR description claims.
3. **`global.json` was added and then removed in `80ce393`.** CI is pinned to 10.0.301 while local builds now follow the ambient SDK, so contributors can hit behavior CI does not reproduce. A `global.json` with `rollForward: latestFeature` would give reproducibility without forcing an exact patch — a trade-off worth making consciously rather than swinging between pinned and unpinned.
4. **Minor:** `#if NET8_0_OR_GREATER` is broader than any current target (`net10.0` is the only modern one). Harmless and forward-looking — and it would become exactly right if suggestion (1) above is taken.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. This is not an independent human review, and no human approval is claimed. Note that this PR modifies shipping code, where AI-only review is a weaker guarantee than it was for the preceding test/CI/docs changes.
**Verification:** Line-by-line review of all `src/` changes; native `net10.0` unit run; `net462` compilation; Release pack with nuspec dependency-group comparison and per-target `project.assets.json` asset-selection parsing; `net10.0` and `net6.0` package consumers; native CLI tool install and `--help`; `Certes.Func` collateral build; branch-protection API read; hosted run 35548024969. Not verified: `net462` runtime execution, live Azure Web App interoperability, CA integration, and behavior of the `netstandard2.0` asset at runtime on .NET 8.

fszlin (commented):
### Review of `80ce3933`

**Verification run locally (macOS ARM64, SDK 10.0.301, clean tree at PR head)**

- `dotnet build src/Certes/Certes.csproj` — all three targets, 0 errors (only the known `docs/docstrap` SourceLink warnings).
- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true` — 145 passed / 0 failed / 0 skipped, native .NET 10, no roll-forward.
- `dotnet pack` (Release, signed) + `dotnet build scripts/PackageSmoke/PackageSmoke.csproj -c Release -f net6.0` — 0 warnings, so `CheckEolTargetFramework` does suppress `NETSDK1138`; `project.assets.json` confirms `net6.0 -> lib/netstandard2.0/Certes.dll` and `net10.0 -> lib/net10.0/Certes.dll`.
- `dotnet run --project scripts/PackageSmoke/PackageSmoke.csproj -c Release -f net10.0` — passes; the asset identity assertion works.
- `dotnet build test/Certes.Tests.Integration/Certes.Tests.Integration.csproj -p:SkipSigning=true` — both `net10.0` and `net462` compile.
- `actionlint` clean, `git diff --check` clean, no `global.json`, all four required checks pass.

**Assessment: correct and self-consistent.**

The `#if NET8_0_OR_GREATER` `[Obsolete(... DiagnosticId = "SYSLIB0051")]` additions are required rather than cosmetic: the `GetObjectData` override would otherwise hit CS0672 under `TreatWarningsAsErrors`, and self-references inside obsolete members suppress SYSLIB0051 without a blanket `NoWarn` (`src/Certes/AcmeRequestException.cs:111`). Targeting is consistent — `IncludeCli` moved to `net10.0` (`test/Certes.Tests/Certes.Tests.csproj:8`) so `X509CertificateLoader` in `test/Certes.Tests/Cli/Commands/AzureAppCommandTests.cs:152` stays net10-only, while the integration test keeps a `#if NET9_0_OR_GREATER` fallback for `net462`. CI reuse is sound: `dotnet run -f net10.0` still restores both TFMs before the compile-only probe.

**Nits (non-blocking)**

1. `scripts/PackageSmoke/Program.cs:11` dereferences `framework` with no null guard; the `#if` means the net6.0 build computes but never uses it. Safe today, just brittle.
2. `docs/ci-migration.md:22` still says hosted verification of this migration is pending, though run 35548024969 passed — stale within this PR.
3. Dropping the `net6.0` asset reroutes .NET 7/8/9 consumers to `netstandard2.0`, which is compile-checked but never executed in CI (nor is `net462`). Runtime dependencies are netstandard2.0-only, so risk is low and the limitation is disclosed, but it is worth an explicit note in release notes alongside the CLI runtime change.
4. `dotnet list ... --vulnerable` for the smoke graph covers only `net10.0`; the `netstandard2.0` consumer graph is not audited.
5. The CLI now requires .NET 10 while the 2.x-style instructions still show a bare `dotnet tool install --global dotnet-certes` (`docs/README.md:127`). Fine while unreleased; needs emphasis in the release notes.

No blockers found. No behavior changes beyond the intended target/runtime change; assembly signing, package identity, and dependency versions are unchanged.

> **AI-assisted review:** Self-review using OpenCode with qwen3.8-flash-next (not independent of the implementation work in this branch).
> **Scope and verification:** Full PR diff (17 files, +137/-81) plus library/test builds, the net10.0 unit suite, Release packing, and both package smoke probes as listed above. Not verified: `net462` runtime tests, .NET 6 consumer execution, hosted ACME/CA interoperability, and live Azure behavior.

fszlin:
Addressed the reviews and SDK-selection request:
- Added a net8.0 library asset, with native .NET 8 package smoke execution in CI alongside .NET 10. CLI and full unit-test baseline remain .NET 10. Documented the November 10, 2026 .NET 8 support deadline.
- Added build-time ReferencePath assertions for expected asset paths and package versions on net10.0, net8.0, and compile-only net6.0. Negative checks confirmed incorrect asset/version expectations fail.
- Made runtime asset inspection null-safe with meaningful expected-framework checks for every consumer target.
- Kept global.json absent as explicitly requested. CI now uses 10.0.x (plus 8.0.x in the package job) and logs dotnet --info in both jobs.
- Promoted the CLI runtime change to a Breaking changes section and documented exact library asset selection and runtime coverage. Recorded prior completed hosted migration verification.

Local verification (SDK 10.0.301, macOS ARM64): 145 native net10.0 unit tests passed; signed library pack passed for all four targets; CLI pack passed; net10.0 consumer executed successfully; net8.0/net6.0 consumers compiled with asset assertions. The machine lacks .NET 8 runtime, so native net8.0 execution is delegated to CI. Fresh-cache NuGet vulnerability audit of the smoke project covered all three consumer targets and reported no vulnerable packages; shipping-project audits also reported none. actionlint, whitespace, and staged secret checks passed. Existing SourceLink warnings/package README notices remain.

AI assistance: OpenCode using gpt-6-astra, implementation and self-review; no independent or human review claimed.

fszlin:
Hosted verification passed at 1d077f5: https://github.com/fszlin/certes/actions/runs/35549055095

10.0.x resolved to SDK 10.0.401 / runtime 10.0.12 on all three OS jobs. Each passed all 145 unit tests. Package checks passed, including native .NET 8 execution with installed runtime 8.0.31, native .NET 10 execution, and the compile-only .NET 6 asset assertion. All four required checks are green.

fszlin (commented):
## Re-review of `1d077f5` ("Add .NET 8 package coverage and enforce asset selection")

All four follow-ups are addressed, and the two substantive ones are addressed well. One change went in the opposite direction from what I raised, which I want to put on the record rather than let pass as "resolved".

| # | Item | Status |
| --- | --- | --- |
| 1 | No `net8.0` asset | Fixed, with native CI execution |
| 2 | CLI runtime break under-emphasized | Fixed — promoted to a **Breaking changes** section |
| 3 | `net6.0` probe asserts nothing | Fixed — MSBuild assertions on all three consumer targets |
| 4 | `#if NET8_0_OR_GREATER` broader than targets | Now exactly correct |
| — | SDK pinning | Changed, but *away* from reproducibility — see below |

### Verification

Clean worktree at `1d077f5`, SDK 10.0.301, macOS ARM64:

- Library builds for all **four** targets: 0 errors. The 4 warnings are one SourceLink `docs/docstrap` notice per target — no new C# warnings from the `net8.0` addition under `TreatWarningsAsErrors`.
- Unit suite natively on `net10.0`: 145 passed, 0 failed.
- Packed and inspected the nupkg: `lib/` contains `net10.0`, `net8.0`, `netstandard2.0`, `net462`, with matching nuspec dependency groups.
- **I confirmed the new assertions are not vacuous**, which matters because an MSBuild item-batched `Error` silently no-ops when the item list is empty. Forcing wrong expectations produces real failures:
  - `net8.0` with a bogus expected path: `Expected lib/bogus/Certes.dll, resolved lib/net8.0/Certes.dll`
  - `net6.0` with a bogus expected path: `Expected lib/bogus/Certes.dll, resolved lib/netstandard2.0/Certes.dll`
  - stale version: `Expected Certes 0.0.0-nope, resolved 0.0.0-rev334b`
  
  So the .NET 8 consumer genuinely binds `lib/net8.0/Certes.dll`, the .NET 6 consumer genuinely falls back to `netstandard2.0`, and CI will now catch a regression in either. The `@(_CertesReference->Count())` guard before the batched checks is the right defensive pattern.
- `net10.0` consumer ran; `net8.0` and `net6.0` compiled with assertions active. I have no .NET 8 runtime locally, so native .NET 8 *execution* is covered only by CI — which run 35549055095 shows passing against installed runtime 8.0.31.
- Hosted run 35549055095 at `1d077f5`: all four required checks green, 145 tests on each OS.

Item 2 is handled better than I suggested: the changelog now separates a **Breaking changes** section and explicitly decouples the assembly version from the package release decision, and `docs/README.md` warns CLI users to install .NET 10 even when their application stays on .NET 8. That is the right place for that warning.

### One correction and one concern

**Correction to the record.** The response says `global.json` was "kept absent as explicitly requested". I did not request that. My point was that the add-then-remove left CI pinned and local builds unpinned, and that the trade-off should be made deliberately — I suggested `rollForward: latestFeature` as a way to get both.

**The concern.** The resolution went the other way: CI moved from `dotnet-version: 10.0.301` to `10.0.x`. Hosted run 35549055095 resolved **SDK 10.0.401 / runtime 10.0.12** — a different *feature band* from the 10.0.3xx used for every prior verification in this series, including the local runs backing this PR's description.

That is defensible, and logging `dotnet --info` plus recording resolved versions is a genuine mitigation. But it is worth naming the specific failure mode it opens in this repository: `misc/certes.props` sets `TreatWarningsAsErrors=True`, so a future 10.0.x SDK that adds an analyzer or obsoletion warning breaks `main` with no repository change, on a schedule nobody controls. Feature-band moves (3xx to 4xx) are exactly where that behavior tends to shift. Combined with local guidance of "install a .NET 10 SDK" (any), a contributor on 10.0.100 and CI on 10.0.401 can legitimately disagree about whether the build passes.

I am not blocking on this — it is a maintenance-policy choice, and floating has real benefits. But "CI is pinned, local is not" has become "neither is pinned", which is not the direction I was pointing, and it should be a conscious decision rather than an artifact of addressing my comment.

**Verdict: approve.** The packaging work is verified end to end and the asset assertions are proven to fail when they should. Only the SDK-floating decision is left to your judgment.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. Not an independent human review; no human approval is claimed. This PR modifies shipping code and packaging, where AI-only review is a weaker guarantee.
**Verification:** Four-target library build and warning inventory; native `net10.0` unit run; pack with `lib/` and nuspec dependency-group inspection; execution of the `net10.0` consumer; compilation of the `net8.0` and `net6.0` consumers; three deliberate negative tests proving the new MSBuild assertions fire; inspection of hosted run 35549055095 including resolved SDK/runtime versions. Not verified: native .NET 8 execution locally (no runtime installed), `net462` runtime behavior, live Azure interoperability, and CA integration.

fszlin:
Human review: I reviewed the changes and verification results. I accept the .NET 10 CLI requirement, the .NET 8 library target, and the decision to use 10.0.x in CI without global.json. Ready to merge.

fszlin:
SDK policy clarification before merge: removing global.json and using 10.0.x were explicit maintainer requests in the implementation conversation, not requests from the reviewer. The maintainer has now instructed merging after discussing the reproducibility trade-off. CI logs the selected SDK/runtime; future SDK updates may require build fixes under warnings-as-errors. This records the decision, not a claim of independent human code review.

PR #335, opened 2026-09-21
Run integration tests against pinned local Pebble

fszlin:
## Summary
- Add Pebble/challtestsrv 2.10.1 Compose infrastructure pinned to multi-platform digests, loopback-only management ports, and a bounded readiness probe.
- Replace hosted Azure/DNS/TLS responder dependencies with local HTTP-01, DNS-01, and TLS-ALPN-01 provisioning/cleanup through challtestsrv. Real validation and Pebble strict mode are enabled.
- Replace broad TLS bypasses with exact server-certificate pinning, hostname/validity checks, fixed loopback endpoints, and disabled proxies/redirects in the integration-only HttpClient. net462 builds but refuses network initialization.
- Add bounded authorization/order polling in test helpers and wait for valid orders before downloading. These are harness changes, not fixes to production polling.
- Preserve the ten existing integration cases and add three missing-response negative cases. Verify exported PFX key/leaf/issuer contents and PEM round-trip; assert DNS SAN identity instead of assuming Pebble copies SANs into CN.
- Add a Linux Pebble integration CI job with failure logs and unconditional teardown. Update agent guidance and document local commands, trust, image pins, and coverage limits.

## Local verification
macOS ARM64, SDK 10.0.301 / runtime 10.0.9, Docker Desktop engine 29.6.2 / Compose 5.3.1:
- 13 integration tests passed, 0 failed, 0 skipped against pinned Pebble, including after fresh stack recreation.
- Full offline unit suite: 149 passed, 0 failed, 0 skipped after adding four TLS-ALPN generator cases.
- Integration/unit projects compile for net10.0 and net462.
- Compose config validation, readiness script, actionlint, shellcheck, git diff --check, and staged Gitleaks scan passed.
- Local containers and network removed after verification.

Hosted run 35554296416 at 3da81dc passed all five jobs, including strict-mode Pebble integration. Existing four required checks retain their names; the new integration job is not yet required.

## Scope and limitations
- No production transport, dependencies, signing, or public API changes; no public CA fallback or OS trust changes.
- Local test CA success is not public CA interoperability. Random nonce rejection/reuse/artificial validation delays are disabled; retry/polling correctness still needs focused production fixes.
- TLS-ALPN certificates are generated by challtestsrv from Certes key authorization; the flow itself does not validate the Certes TlsAlpnCertificate generator; four new offline cases verify its keys, self-signature, SAN, and critical ACME digest extension.
- The harness uses fixed ports/subnet and one test process per stack. HTTP requests have 10-second timeouts and polling is bounded to 60 seconds. The Generate example explicitly uses 60 retries.
- Server certificate expires October 3, 2027; update it together with the image pins. Public certificate only, no private key added.
- Podman and net462 runtime execution are not verified. The unused net7.0 Functions helper is retained pending a separate retirement change.

## AI assistance
Implemented and self-reviewed with OpenCode using gpt-6-astra. AI assisted with implementation and verification; no independent review or human approval is claimed.


fszlin:
Hosted verification passed at 0bc5579: https://github.com/fszlin/certes/actions/runs/35550345812

- Pebble integration on Linux: 13 passed, 0 failed, 0 skipped; container teardown succeeded.
- Full offline unit suite on Linux, Windows, and macOS: all three jobs green.

This adds verified Linux Docker coverage alongside local Docker Desktop/macOS ARM64. Public CA interoperability, Podman, and net462 runtime execution remain unverified. The integration check is not yet required by branch protection.

AI-assisted verification: OpenCode using gpt-6-astra, self-review only.

fszlin (commented):
## Review: local Pebble integration harness

**Verdict: approve.** I ran the harness end to end on a clean worktree with real containers, and separately proved the new TLS pinning is not vacuous. The security posture is a clear improvement over the blanket `DangerousAcceptAnyServerCertificateValidator` it replaces. One product API lost its only test, which is disclosed but worth closing cheaply.

### Verification performed

Clean worktree at `0bc5579`, macOS ARM64, Docker Desktop 29.6.2, SDK 10.0.301:

- `docker compose up -d` + `wait.sh` from a cold start: ready on first attempt.
- **Integration suite: 13 passed, 0 failed, 0 skipped**, against the pinned containers. Reproduced the stated result on an independent machine.
- Offline unit suite unaffected; `net462` integration compilation: 0 errors.
- Full teardown verified; no containers or networks left behind.

**The pinning is real, not decorative.** I replaced the copied `Pebble/localhost.pem` in the test output with an unrelated self-signed `CN=localhost` certificate and re-ran: every test failed to connect. So `expected.SequenceEqual(certificate.RawData)` genuinely gates the connection, and a MITM or substituted server is rejected. Restoring the real certificate returned the suite to 13/13.

Reviewing the callback itself: exact byte-comparison of the leaf is the decisive check; host and port are constrained to loopback 14000/15000; `RemoteCertificateNameMismatch` and `NotAvailable` must be absent, so hostname validation still applies; validity is checked explicitly with correct `ToUniversalTime()` conversion; and `RemoteCertificateChainErrors` is deliberately ignored, which is right for a pinned self-signed certificate. The whole thing lives in the integration project, and `net462` throws `PlatformNotSupportedException` rather than silently falling back. That satisfies the AGENTS.md rule about TLS bypasses never reaching production transport.

Other checks: `scripts/Pebble/localhost.pem` contains exactly one certificate and **no private key** (`CN=localhost`, SAN `localhost`/`pebble`/`127.0.0.1`, expires 2027-10-03, matching the description). Both images are digest-pinned. Ports bind `127.0.0.1` only. The CI job has a timeout, failure-only container logs, and `if: always()` teardown. Hosted checks on this PR are green, including `Pebble integration`.

### Findings

1. **`IKey.TlsAlpnCertificate` now has zero test references anywhere in the repository.** I grepped `src/` and `test/` to confirm. The old TLS-ALPN test generated the validation certificate with Certes and served it; the new one delegates generation to challtestsrv, so this public API — which produces a cryptographic artifact with the `id-pe-acmeIdentifier` extension — is no longer exercised at all. The description discloses this honestly, but "not validated by this flow" understates it: nothing validates it now.

   This is cheap to close **offline**, without Pebble: generate the certificate and assert it is self-signed, carries the subject name as a DNS SAN, and that the critical `acmeIdentifier` extension contains the SHA-256 of the key authorization. That would be a stronger check than the old integration test, which only proved Pebble accepted it.

2. **A pinning or expiry failure is reported as "Local Pebble is not ready".** In my negative test above, a certificate mismatch surfaced as `Local Pebble is not ready. Run docker compose -f scripts/Pebble/compose.yml up -d`. The real cause is preserved as the inner exception, so it is discoverable, but the headline actively misdirects. This matters on a known date: when the pinned certificate expires on **2027-10-03**, every developer will be told to start containers that are already running. Distinguishing TLS/pin failures from connection failures in `Initialize()` would save that debugging session.

3. **The harness disables exactly the conditions the roadmap says are broken.** `PEBBLE_VA_NOSLEEP=1`, `PEBBLE_WFE_NONCEREJECT=0`, and `PEBBLE_AUTHZREUSE=0` remove validation delays, random nonce rejection, and authorization reuse. That is the right call for a deterministic first landing, and it is disclosed. But AGENTS.md lists discarded `Retry-After`, single-retry `Generate()`, and bad-nonce handling as open production defects, and this configuration guarantees they are never hit. Consider a second, non-required job with chaos enabled once the production polling fixes land — otherwise the suite will stay green precisely because it avoids the known weaknesses.

4. **`-strict=false` is passed explicitly.** If that is only restating Pebble's default, fine; if strict mode would surface client-side protocol deviations, enabling it is a cheap way to make this harness test conformance rather than just happy-path issuance. Worth a deliberate answer either way.

5. **The `Pebble integration` job is not a required check.** It is now the only thing exercising the ACME protocol end to end, so it can regress silently. Adding it to branch protection is the natural follow-up, ideally after a few runs establish its flakiness profile — networked container jobs deserve that caution before becoming merge-blocking.

### Coverage direction

Net positive overall, separate from finding 1: the ten preserved cases now assert exported PFX key/leaf/issuer contents and a PEM round-trip through `AssertExport`, where several previously built a PFX and discarded it; three genuine negative cases were added for missing challenge responses across all three types; and bounded polling replaces unbounded `while (true)` loops, so a broken flow now fails with a timeout instead of hanging CI.

---

**Verification:** Cold-start container bring-up and readiness probe; full 13-test integration run against real Pebble/challtestsrv; a deliberate certificate-substitution test proving the pin rejects non-pinned servers; `net462` compilation; inspection of the pinned PEM for private-key material, subject, SAN, and expiry; review of compose digests, port bindings, and the CI job's diagnostics/teardown; repository-wide grep for `TlsAlpnCertificate` references; confirmation of full teardown. Not verified: Podman, `net462` runtime behavior, public CA interoperability, and behavior under Pebble's chaos options (disabled by this configuration).

fszlin (commented):
### Review of `0bc5579`

**Verified locally (macOS ARM64, SDK 10.0.301, Docker 29.6.2 / Compose 5.3.1, clean tree at this head)**

- Stack starts and `scripts/Pebble/wait.sh` exits 0; the pinned digests resolve on arm64 (multi-arch manifest confirmed); all three published host ports bind to `127.0.0.1` only.
- The certificate served on `localhost:14000` matches the committed `scripts/Pebble/localhost.pem` (SHA-256 fingerprint `AC:25:E1:9D:...:1F:E9`). The committed file is public-key only (no private key), SAN `localhost` / `pebble` / `127.0.0.1`, valid 2025-09-03 to 2027-10-03.
- Integration suite: **13 passed / 0 failed / 0 skipped** on `net10.0`. Offline unit suite still **145 passed / 0 failed / 0 skipped**. Both test projects also compile for `net462` (only the known `docs/docstrap` SourceLink warning).
- **Validation is real, not bypassed.** Container logs show HTTP-01 probes to `:5002`, the TLS-ALPN key authorization received correctly by challtestsrv (so `content` is the correct payload field for `add-tlsalpn01`), per-host TXT records added and removed, wildcard identifiers normalized by challtestsrv to `_acme-challenge.wildcard.example.test.` (the helper's literal `*.` prefix is accepted), and the negative cases ending `INVALID`.
- `actionlint`, `shellcheck scripts/Pebble/wait.sh`, `docker compose config --quiet`, and `git diff --check` all clean. Containers and the `certes-pebble_acmenet` network were removed after verification.

**Assessment: approve.**

The security posture is a clear improvement over what it replaces: `DangerousAcceptAnyServerCertificateValidator` and the process-wide static `ServicePointManager.ServerCertificateValidationCallback` are gone, and the pin is scoped to `localhost` on ports 14000/15000. Skipping `chain.ChainPolicyResult` is correct here because the callback compares the exact `RawData` of the pinned certificate, so a rogue-but-valid chain cannot satisfy it. In particular it does not inherit the `PfxBuilder` chain-building weakness noted in the revival baseline. The commit itself is also safe: `scripts/Pebble/localhost.pem` is already inside the pinned container image, so publishing it leaks nothing, and no key material was added.

**Findings**

1. `test/Certes.Tests.Integration/Readme.cs:57` — `retryCount: 60` diverges from the documented example (`docs/README.md:96` uses the default `retryCount = 1`, see `src/Certes/Extensions/IOrderContextExtensions.cs:70`). This is the only test that exercises `Generate()`, and `PEBBLE_VA_NOSLEEP=1` removes exactly the processing latency that exposes the still-unfixed polling defect, so the documented flow is left with no coverage. Either add a traceability comment tying the override to the known `RetryAfter` issue, or keep the default and let the failure surface.
2. `.editorconfig:15` requires `charset = utf-8-bom`, but `IntegrationHelper.cs`, `AcmeContextIntegration.CertificateByTlsAlpnTests.cs` and the new `ChallengeFailureTests.cs` have no BOM (sibling integration files still do).
3. `Initialize()`'s blanket `catch` relabels every failure as "Local Pebble is not ready", which would misreport a future management-API or account-provisioning break as a missing container stack.
4. `AddTestCerts()` depends on the static `root` populated only by `GetAcmeUriV2()`, so calling it first yields an opaque failure rather than a clear message.
5. `ChallengeFailureTests.cs:16` mixes `[InlineData(ChallengeTypes.Http01)]` with the literal `"tls-alpn-01"` because `ChallengeTypes.TlsAlpn01` is a property rather than a `const` (`src/Certes/Acme/Resource/ChallengeTypes.cs:24`). Pre-existing, and converting it is binary-breaking, so worth a separate follow-up rather than a change here.
6. `scripts/Pebble/README.md:33` — "Missing-response tests for each challenge type ensure validation is not bypassed" overstates it: those cases prove the negative path only. Real-validation assurance comes from the issuance tests plus `PEBBLE_AUTHZREUSE=0` (I confirmed this from the container logs rather than from the claim).
7. `Pebble integration` is not a required check, so a broken harness will not block merges. Correct while the job is new, but `docs/ci-migration.md:48` should carry a follow-up to promote it once stable.

No production code, signing, package identity, or dependency versions touched. Not verified: `net462` runtime execution, Podman, and public-CA interoperability.

---

> **AI-assisted review:** Self-review using OpenCode with qwen3.8-flash-next; the same harness verified the branch locally but did not author it. This is not an independent human review and no human approval is claimed.
> **Scope and verification:** Full PR diff (17 files, +440/-294) reviewed, plus a live local run of the pinned stack, the 13 integration tests, the 145 offline unit tests, `net462` compilation of both test projects, container-log inspection to confirm genuine challenge validation, certificate fingerprint comparison against the committed pin, `actionlint`, `shellcheck`, Compose config validation, and `git diff --check`. Not verified: `net462` runtime execution, Podman, hosted run of the new `Pebble integration` job, and public-CA interoperability.

fszlin:
Addressed both reviews:
- Added four offline TlsAlpnCertificate cases (RSA, ES256/384/512): matching public key, self-signature, DNS SAN, critical ACME identifier OID, and DER-decoded SHA-256 key-authorization digest. Full unit suite is now 149 cases.
- Initialization errors now name the failing stage; .NET 10 distinguishes TLS/pin failures from connection failures. Other HTTP/management/account errors preserve their inner exception without blaming missing containers. Issuance-root use before successful initialization now fails clearly.
- Enabled Pebble strict mode after checking its v2.10.1 source and verifying all 13 integration tests pass. It rejects legacy resource fields and challenge responses other than the empty JSON object.
- Added traceability comments for the Generate retry override and the TLS-ALPN InlineData literal; the public property remains unchanged.
- Restored UTF-8 BOMs on the three flagged C# files.
- Narrowed validation claims and documented offline generator coverage, future stress settings, and promotion of Pebble integration to required status after stable strict-mode runs.

Verification: 149 unit tests and 13 strict-mode integration tests passed locally (SDK 10.0.301 / .NET 10.0.9, macOS ARM64/Docker Desktop). Both test targets compile. A stopped-stack negative run confirmed the actionable connection-stage diagnostic. actionlint, shellcheck, whitespace and staged secret checks passed. Stack torn down afterward. TLS mismatch diagnostic classification was inspected but no new substituted-pin test was performed in this session.

AI assistance: OpenCode using gpt-6-astra, implementation and self-review. No independent or human review claimed.

fszlin:
Hosted verification passed at 3da81dc: https://github.com/fszlin/certes/actions/runs/35554296416 — all three OS build/unit-test jobs, package smoke checks, and strict-mode Pebble integration are green.

fszlin (commented):
## Re-review of `3da81dc` ("Address Pebble review with TLS-ALPN coverage and diagnostics")

All five items are addressed, two of them beyond what I asked. I re-ran the harness against real containers and specifically performed the two verifications the response did **not** claim to have done.

| # | Item | Status |
| --- | --- | --- |
| 1 | `TlsAlpnCertificate` untested | Fixed — 4 offline cases, mutation-tested below |
| 2 | Pin/expiry failures blamed missing containers | Fixed — verified by substituted-pin test |
| 3 | Chaos options disable the known-broken paths | Documented, with a planned stress job |
| 4 | `-strict=false` | **Enabled** `-strict=true`, verified passing |
| 5 | Integration job not required | Promotion path documented |

### Verification

Clean worktree at `3da81dc`, macOS ARM64, Docker Desktop 29.6.2, SDK 10.0.301:

- Unit suite: **149 passed, 0 failed** (145 + the 4 new TLS-ALPN cases).
- Integration suite under **strict mode**: **13 passed, 0 failed** against freshly started pinned containers. Strict mode is a real behavior change to the harness, so I treated it as new rather than assuming the previous run carried over.
- Full teardown confirmed; no containers or networks left behind.

**The substituted-pin test you explicitly did not repeat.** The response notes "TLS mismatch diagnostic classification was inspected but no new substituted-pin test was performed", so I ran it. Replacing the copied `localhost.pem` with an unrelated self-signed `CN=localhost` certificate now produces:

> `Pebble TLS validation failed while fetching the ACME directory...`

and with the stack stopped:

> `Could not connect to local Pebble while fetching the ACME directory...`

So the classification is correct in both directions on real failures, not just by inspection. The 2027-10-03 expiry will now report a TLS problem instead of telling someone to start containers that are already running.

**The new TLS-ALPN test is not vacuous.** Digest assertions are easy to write in a way that passes regardless of input, so I mutated the expected value to `KeyAuthorization(token + "MUTANT")`: all four cases failed, as they should. The test genuinely pins the RFC 8737 §3 payload — DER `OCTET STRING` of the SHA-256 key authorization — along with the critical `1.3.6.1.5.5.7.1.31` OID, DNS SAN, self-signature, and public-key correspondence, across RSA and all three ECDSA curves. That is materially stronger than the integration coverage it replaces, which only proved Pebble accepted the certificate.

Hosted run 35554296416 at `3da81dc`: all five checks green, including strict-mode `Pebble integration`.

### Notes on the responses

Enabling `-strict=true` goes further than my question, which only asked for a deliberate answer. Checking Pebble's v2.10.1 source to establish what strict mode rejects (legacy ACME v1 resource fields, non-empty challenge responses) and confirming the suite still passes is the right way to make that change rather than flipping the flag hopefully.

Two unprompted improvements worth noting: `root` is now assigned only after account provisioning succeeds, so a partial failure can no longer leave a half-initialized static; and `AddTestCerts` throws a clear error instead of `NullReferenceException` when used before initialization. Both are the kind of latent-state bug that would have produced confusing failures later.

Item 3's resolution is documentation plus a planned stress job, which I think is correct sequencing — enabling nonce rejection and validation delays before the production polling fixes land would produce a flaky suite that proves nothing. The commitment is recorded in both `scripts/Pebble/README.md` and the replacement plan, and the README is appropriately explicit that negative cases "alone do not establish complete protocol conformance".

**Verdict: approve.** No outstanding items.

---

**Verification:** 149-case unit run; 13-case strict-mode integration run against real Pebble/challtestsrv containers from a cold start; substituted-pin negative test and stopped-stack negative test confirming both new diagnostic paths; mutation test of the new TLS-ALPN digest assertion; teardown confirmation; inspection of hosted run 35554296416. Not verified: Podman, `net462` runtime behavior, public CA interoperability, and behavior under Pebble's still-disabled chaos options.

fszlin:
Merged as d52b764 after the final re-review and all five passing checks. Pebble integration is now required on main alongside the three Build OS checks and Package smoke checks, restricted to the GitHub Actions app. Strict/up-to-date checks and all existing required contexts were preserved. This supersedes the pre-merge migration notes stating the integration check was not yet required; those notes should be refreshed in the next documentation change.

AI-assisted merge/settings verification: OpenCode using gpt-6-astra, performed on explicit maintainer instruction.

PR #336, opened 2026-09-21
Fix order polling and Retry-After handling

fszlin:
## Summary
- propagate ACME response `Retry-After` metadata through resource and finalize responses
- poll pending and processing orders using server-directed delays, with a one-second minimum and a 15-minute per-delay safety cap
- preserve explicit caller retry budgets while defaulting new callers to 60 retries
- handle stale HTTP-date `Retry-After` values as immediate retries
- add regression coverage and update polling documentation

The OpenCode configuration change was reverted from this branch and moved to PR #337.

## Verification
- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true`: 152 passed, 0 failed, 0 skipped
- `git diff --check`: passed
- Hosted build, package smoke, and Pebble integration checks: passed

**AI assistance:** Prepared with OpenCode using gpt-5.6-luna. AI assisted with implementation, review, and PR maintenance. No human review is claimed by this disclosure.


fszlin (commented):
## Review of 93a0166

**Verdict: changes recommended before merge.** Response metadata propagation is correct and useful, and the delay injection makes tests fast. However, the new hard caps undermine the server-directed timing and change an existing caller-controlled API contract.

### Findings

1. **[P1] Do not shorten a server Retry-After value to 60 seconds.** `src/Certes/Extensions/IOrderContextExtensions.cs:97` uses `Math.Min(..., 60)`, so a legitimate `Retry-After: 120` makes the client poll after 60 seconds. RFC 8555 section 7.4 explicitly directs clients in processing state to send POST-as-GET after the time given in Retry-After. Polling early adds unnecessary requests and can provoke rate limiting. The new `PollsPendingAndProcessingUsingBoundedRetryAfterDelay` test actually locks in this behavior by expecting 60 for a server value of 120. Honor positive server delays; if they exceed an explicit client time budget, terminate with a clear timeout instead of sending an early poll. A finite retry count already bounds the number of requests.

2. **[P2] Preserve explicit retryCount values instead of silently truncating them.** `IOrderContextExtensions.cs:94` clamps all values to MaxRetryCount=60. An existing caller supplying 100 retries previously could reach a valid order on poll 61; after this PR it throws before that point. A default of 60 is reasonable, but a finite caller-supplied budget is already bounded. Remove the hidden maximum, or make any replacement limit an explicit, documented API decision. Add a case that becomes valid after more than 60 polls with a larger supplied budget; the current BoundsProcessingPolling test endorses the incompatible truncation rather than protecting the existing contract.

3. **[P2] Document the public polling behavior and update stale baseline claims.** This is not purely internal: Generate is public and its default changes from 1 to 60, significantly changing latency on slow/failed issuance. Update the changelog/usage guidance and AGENTS.md: its known-issues section still says Resource/Finalize discard RetryAfter and Generate defaults to one retry. The integration Readme retry override/comment also refers to the old behavior. Note that C# optional argument defaults are baked into callers: already-compiled callers continue passing 1 until recompiled; replacing the DLL alone does not give them the new default.

### Additional coverage worth adding
- Verify changing RetryAfter values across successive resource responses, plus reset to zero when the header disappears. Separate propagation tests and a constant mocked value do not cover that end-to-end behavior.
- Exercise initial pending -> ready before finalize. The existing code still finalizes pending immediately; RFC 8555 says finalize before ready produces orderNotReady. The new test covers processing -> pending *after* finalize, not that real-world sequence. This is pre-existing behavior rather than a regression introduced here, but the PR should not imply pending-order orchestration is fully fixed.
- Cover network errors, terminal invalid status, and zero retry budget without downloading; the existing invalid test does not verify Download is never called.

### Verification
- Full unit suite on macOS ARM64, SDK 10.0.301 / runtime 10.0.9: **152 passed, 0 failed, 0 skipped**.
- Library build passed for net10.0, net8.0, netstandard2.0, net462.
- Integration project net462 compilation passed.
- git diff --check passed.
- All five hosted checks at the reviewed head are green, including Pebble integration. I inspected check results but did not rerun local Pebble or package smoke.
- Known SourceLink documentation-submodule warnings remain. No net462 runtime or public-CA interoperability claim.

Reference: https://www.rfc-editor.org/rfc/rfc8555#section-7.4

**AI-assisted review:** OpenCode using gpt-6-astra. Review of changes authored in another session; not an independent human review. No source changes made; unrelated untracked OpenCode configuration was preserved.

fszlin (commented):
## Review: order polling and Retry-After handling

**Verdict: approve the fix, but two things should change before merge** — the changelog claim, and the undocumented `retryCount` cap. The underlying diagnosis is correct and the tests are well targeted.

### The fix is real, and larger than it looks

I checked whether `RetryAfter` was ever populated before this change: on `main`, **nothing assigns `EntityContext.RetryAfter` anywhere in `src/`**. It stayed `0` for the entire lifetime of every context, so `Math.Max(context.RetryAfter, 1)` always evaluated to 1 second, and `ExtractRetryAfterHeaderFromResponse` — including its HTTP-date handling — was unreachable dead code. This PR is what makes the feature exist at all, not merely a tuning change.

The mechanics are sound: `RetryAfter` has a `protected` setter and `EntityContext<T>` is internal, `IResourceContext.RetryAfter` stays get-only, and the new `delay` seam is a non-extension `internal` overload, so **no public API surface changes**. Adding `Pending` to the poll loop is correct for pre-draft-11 servers, which the pre-finalize guard already treats as equivalent to `Ready`.

Verified: 152 unit tests pass on `net10.0`; all five required checks green. The new tests assert the right things, and `BoundsProcessingPolling`'s expectation of 62 `Resource()` calls is arithmetically correct once you account for `CreateCsr` calling `Resource()` a second time.

### 1. Worst-case blocking is now one hour, and it cannot be cancelled

I probed the actual bound using the new internal seam, with a server asking for a long `Retry-After` and a caller passing `int.MaxValue`:

```
polls=60  totalWait=01:00:00  resourceCalls=62
```

The default went from **1 retry to 60**, and the per-wait cap of 60s means `Generate` can now block for a full hour. `Generate` takes no `CancellationToken`, and AGENTS.md already lists cancellation as unimplemented — so this change multiplies the cost of that gap by sixty. Before, a stuck order failed in about a second.

This is defensible as the intended behavior (polling until valid is what callers want), but it is an observable change in a public API's default, so:

**The "Documentation update is not necessary" checkbox is wrong.** This belongs in `docs/CHANGELOG.md` under Changed, stating the new default and the worst-case bound. A consumer calling `Generate` inside a request handler or a timed job needs to know it can now block for an hour where it previously returned.

### 2. `retryCount` is silently clamped, and the doc does not say so

`Math.Min(Math.Max(retryCount, 0), MaxRetryCount)` caps any caller-supplied value at 60. `BoundsProcessingPolling` passes 100 and asserts exactly 60 polls, so this is deliberate — but the XML doc reads "Maximum number of polling retries while the Order is pending or processing. (default = 60)" with no mention that larger values are ignored. Someone passing 300 after a CA incident will get 60 and no indication why.

Either document the cap in the `<param>` text, or honour the caller and bound total elapsed time instead. Silently discarding an explicit argument is the one thing I would not ship as-is.

### 3. A pre-existing bug just became reachable

```csharp
if (date.HasValue)
    return Math.Abs((date.Value - DateTime.UtcNow).TotalSeconds);
```

`Math.Abs` turns a **past** `Retry-After` date into a positive wait. With clock skew, or a server echoing a stale date, "retry immediately" becomes "wait", and under this PR's 60s cap a date 10 minutes stale produces the maximum 60-second delay on every single poll — 60 times.

This was harmless when `RetryAfter` was dead. It is live now, and this PR is specifically about `Retry-After` correctness, so it is the natural place to clamp negatives to zero rather than mirroring them. If you would rather keep the change minimal, it deserves an issue rather than silence.

### 4. Capping the wait means polling faster than the server asked

A CA under load that sends `Retry-After: 300` will now be polled every 60 seconds instead — five times more often, precisely when it is least able to absorb it, and precisely when rate limits are most likely to trigger. A client-side bound is reasonable, but bounding *total elapsed time* while honouring the server's requested interval achieves termination without overriding the server. Worth a deliberate decision.

### Minor

The AI disclosure says "using the current session model". AGENTS.md asks for the harness and model when known, and not to guess otherwise — "the current session model" is neither a name nor an admission that it is unknown.

---

**AI-assisted review:** Performed with OpenCode using claude-opus-5, in a separate session from the one that authored the PR. Not an independent human review; no human approval is claimed. This PR changes shipping behavior, where AI-only review is a weaker guarantee.
**Verification:** Full `net10.0` unit run (152 passed); `git grep` across `origin/main` confirming `RetryAfter` was never assigned before this change; reading of `CreateCsr` to confirm the 62-call expectation; an ad-hoc probe through the new internal seam measuring the real worst-case polling budget (60 polls, 1 hour, `int.MaxValue` clamped to 60); review of public API surface impact; inspection of the five green checks. Not verified: behavior against a real CA that emits `Retry-After` (the Pebble harness disables validation delays), `net462` runtime behavior, and clock-skew handling end to end.

fszlin:
Addressed the actionable review feedback in commit `5f30378`.

Changes:
- Removed the 60-second cap on server-provided `Retry-After` delays.
- Preserved explicit `retryCount` values above the default of 60.
- Clamped stale HTTP-date `Retry-After` values to zero instead of turning them into a positive delay.
- Added coverage for changing `Retry-After` values, including reset to the one-second minimum when the header disappears.
- Updated the polling budget test to verify that a caller-supplied budget of 100 is honored.
- Updated the changelog, `AGENTS.md`, Pebble guidance, and integration comments to document the new polling behavior and optional-argument default caveat.

Verification: `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true` passed with 152 tests, 0 failures, and 0 skips. `git diff --check` also passed.

The pre-existing pending-before-finalize orchestration, cancellation support, and broader network/terminal-state coverage remain outside this review fix.

fszlin (commented):
## Re-review of 5f30378

**The three findings from my previous review are addressed. No remaining blocker from that review.**

1. Server delays are no longer capped: Generate uses Math.Max(context.RetryAfter, 1). The regression test now expects 120 seconds for RetryAfter=120 and the 1-second fallback for the next zero value.
2. Explicit retry budgets are preserved: the hidden maximum was removed. HonorsCallerPollingBudget supplies 100 and verifies 102 resource calls (two preparatory reads plus 100 polls), rather than the former 62.
3. The changelog documents the new 60-retry default, longer possible waits, explicit budgets, and the need to recompile callers to pick up changed C# optional defaults. AGENTS.md and the integration example no longer claim that metadata is discarded or the default is one.

The additional HTTP-date change from Math.Abs to Math.Max(..., 0) is directionally correct: expired Retry-After dates must not become future delays.

### Remaining non-blocking follow-ups
- The PR description is stale: it still claims hard caps of 60 retries/60 seconds, delay-capping coverage, and that documentation was unnecessary. Please align it with the updated code and changelog.
- The revised test varies a mocked RetryAfter property; it does not verify a real EntityContext resetting its stored value when a subsequent response omits the header. Header-date past/future cases also merit focused regression coverage.
- Initial pending -> ready orchestration remains unchanged: Generate still attempts finalize on an initial pending order. The added polling is post-finalize only. Keep that limitation distinct from the fixes verified here.
- The integration example comment says its count is bounded independently of server delays; clarify that the *number of retries*, not elapsed time, is bounded. Existing test-helper 60-second deadlines also remain, despite their removal from the Pebble README wording.
- Commit 1ee450c adds OpenCode configuration/.gitignore changes unrelated to polling. Prefer a separate focused PR, or explicitly acknowledge the extra scope. I did not validate plugin behavior in this review.

### Verification
Re-ran the full net10.0 unit suite at 5f30378 on macOS ARM64, SDK 10.0.301 / runtime 10.0.9: **152 passed, 0 failed, 0 skipped**. git diff --check passed. All five hosted checks at this head passed, including package smoke and Pebble integration. Known SourceLink warnings remain. Local integration tests and all-target builds were not repeated for this follow-up; no net462 runtime or public-CA interoperability claim.

**AI-assisted review:** OpenCode using gpt-6-astra; review of changes authored in another session, not an independent human review. No source changes made.

fszlin (commented):
## Re-review of `5f30378` ("Address order polling review feedback")

Three of the four items are fully fixed, and the changelog entry is better than what I asked for. But **fixing item 4 the way it was fixed made item 1 materially worse**, and I do not think it should merge in this state.

| # | Item | Status |
| --- | --- | --- |
| 1 | Uncancellable blocking, undocumented | Documented — but the bound is now *unbounded* (below) |
| 2 | `retryCount` silently clamped | Fixed |
| 3 | `Math.Abs` on stale `Retry-After` dates | Fixed |
| 4 | Per-wait cap overrides the server | "Fixed" by removing the cap entirely — see below |

### What is genuinely fixed

- **Item 2.** `Math.Max(retryCount, 0)` now preserves caller budgets; `HonorsCallerPollingBudget` asserts 102 `Resource()` calls for a budget of 100. The `<param>` doc records the negative-to-zero behavior.
- **Item 3.** `Math.Max((date - UtcNow).TotalSeconds, 0)` is the right correction; a stale date now falls through to the one-second minimum.
- **Item 1's documentation half.** The changelog entry is accurate and goes further than I suggested: flagging that **C# embeds optional-argument defaults at the call site**, so already-compiled consumers keep `retryCount = 1` until they recompile, is a sharp observation I had not made. That materially changes who is affected and when.
- Test updates are coherent: `PollsPendingAndProcessingUsingServerRetryAfterDelay` now proves a 120-second server value is honored *and* that a disappearing header falls back to one second. AGENTS.md's baseline correctly drops the now-fixed defects.
- 152 tests pass on `net10.0`; all five checks green.

### The problem: blocking is now unbounded and server-controlled

Removing `MaxRetryAfterSeconds` honours the server, which was the goal — but nothing replaced it, so there is no longer any ceiling on how long `Generate` can block. I measured it through the internal seam with a hostile header value:

```
singleWait=24855.03:14:07   (≈ 68 years, in ONE await)
polls=3  total=74565.09:42:21
```

A single `Retry-After: 2147483647` now hangs the calling thread effectively forever, and `Generate` still takes no `CancellationToken`. The realistic case is bad too: a CA under incident emitting `Retry-After: 86400` gives a **60-day** budget at the new default of 60 retries. Before this PR the worst case was one hour; before the whole series it was about one second.

My earlier suggestion was to *"bound total elapsed time while honouring the server's requested interval"*. That is two halves, and only the first was implemented. Honouring an untrusted, server-supplied duration with no overall deadline and no cancellation is the more dangerous half to ship on its own — it converts a response header into unbounded client-side availability impact.

Any one of these would resolve it:
- an overall deadline (for example `TimeSpan maxWait`) checked across the whole loop, which is the shape I meant;
- a generous per-wait ceiling — 10 or 15 minutes respects every real CA I am aware of while keeping the worst case finite;
- a `CancellationToken` overload, which is on the roadmap anyway and would let callers impose their own bound.

The existing tests would not catch this, because they inject a no-op delay: `PollsPendingAndProcessingUsingServerRetryAfterDelay` asserts the *requested* `TimeSpan`, never that it is bounded. A test asserting a maximum requested delay would lock in whichever fix you choose.

### Process: unrelated commit in a behavior-fix PR

`1ee450c` "Add OpenCode repository configuration" adds `opencode.jsonc` and a `.gitignore` entry. That is editor tooling in a PR whose stated scope is ACME polling, and AGENTS.md is explicit: *"Keep each change focused. Separate dependency migrations, behavior fixes, and broad formatting changes so their effects can be verified independently."*

It also commits a third-party plugin (`open-codebase-index`) that every contributor using OpenCode in this repository will load automatically. That deserves its own PR and its own decision, not a drive-by in an order-polling fix. It does not affect the shipped library, so it is a process objection rather than a correctness one — but it is exactly the kind of mixing the guide asks to avoid.

### Verdict

**Request changes** on the unbounded-wait regression; the rest is good. Items 2 and 3 are correct, and the changelog is genuinely better than what I asked for.

---

**Verification:** Full `net10.0` unit run (152 passed); source review of the clamp removal, `Math.Max` date correction, and doc changes; an ad-hoc probe through the internal `delay` seam measuring a single `Retry-After: int.MaxValue` await at ~24,855 days; review of the updated tests for what they do and do not assert; inspection of `1ee450c`'s contents and the five green checks. Not verified: behavior against a real CA emitting `Retry-After` (the Pebble harness disables validation delays), and `net462` runtime behavior.

fszlin:
Addressed the remaining blocking review feedback in commit `848ed8b`.

- Added a 15-minute per-delay safety cap for server-directed `Retry-After` values, preventing a malformed or abusive header from causing an effectively unbounded await.
- Preserved normal server-directed delays below the cap, including the existing 120-second case.
- Extended the regression test to verify the 120-second delay, one-second fallback when the header disappears, and capping of `int.MaxValue` to 15 minutes.
- Clarified that the integration example bounds the number of retries, not elapsed time.
- Updated the changelog and Pebble documentation with the safety cap.
- The unrelated OpenCode configuration commit was already reverted from this PR and is isolated in PR #337.

Verification: `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true` passed with 152 tests, 0 failures, and 0 skips. `git diff --check` passed. The local untracked `.opencode/` directory was not included.

fszlin (commented):
## Follow-up review of 848ed8b

**One original issue has been reintroduced; changes recommended before merge.** This supersedes my no-remaining-blocker conclusion for 5f30378.

**[P1] Server-directed delays are shortened again — src/Certes/Extensions/IOrderContextExtensions.cs:95.** The new Math.Min(..., MaxRetryAfterSeconds) caps every wait at 900 seconds. For a valid Retry-After: 1800, Generate sends the next poll after 900 seconds, before the server-requested time. Raising the old cap from 60 seconds to 15 minutes changes the threshold but does not resolve the original issue (RFC 8555 section 7.4). A client wait limit is reasonable, including avoiding Task.Delay range errors for extreme values, but exceeding that limit should fail clearly without sending an early poll. Alternatively, honor the interval using supported wait chunks. Please test an ordinary above-limit value such as 1800 and assert either the full wait or failure without another Resource call; the current int.MaxValue test endorses early polling.

Other findings:
- Explicit caller retry budgets remain preserved.
- Changelog/default/recompilation documentation remains in place, though the new cap is now documented and AGENTS.md still claims server intervals are honored without qualification.
- The integration comment now correctly distinguishes retry count from elapsed time.
- Unrelated OpenCode configuration changes are reverted in the net PR diff.
- The PR description remains stale: it still claims 60-retry/60-second hard caps and no documentation update.
- The previously noted header-reset/date test gaps and initial pending-before-finalize limitation remain.

**Verification:** At 848ed8b, re-ran the full net10.0 unit suite on macOS ARM64: 152 passed, 0 failed, 0 skipped. git diff --check passed. Hosted Linux, macOS, package smoke and Pebble checks passed; Windows was still pending at the last check. Known SourceLink warnings remain. No local integration rerun or public-CA interoperability verification.

**AI-assisted review:** OpenCode using gpt-6-astra; review of changes authored in another session, not an independent human review. No source changes made.

fszlin (commented):
## Re-review of `848ed8b` ("Bound server-directed polling delays")

Resolved. The 15-minute per-delay cap is one of the three options I proposed, and it is the one that keeps real-CA behavior intact while making the worst case finite.

### Verification

I re-ran the probe that produced the 68-year await, now against the default budget of 60 retries:

```
retryAfter=900          single=00:15:00  polls=60  total=15:00:00
retryAfter=2147483647   single=00:15:00  polls=60  total=15:00:00
```

A hostile `Retry-After: 2147483647` is clamped to 15 minutes, so the header no longer controls how long the client blocks. Worst case is now a finite 15 hours, against an unbounded await before this commit. Normal server values pass through untouched — the 120-second case is still exercised.

The regression test now asserts the bound directly (`delays[2] == TimeSpan.FromMinutes(15)` alongside the 120-second and one-second cases), which is the missing assertion I flagged. That locks the behavior in; the previous tests would have silently accepted any duration.

152 tests pass on `net10.0`; all five checks green on `848ed8b`.

### Scope

`3ab0c3f` reverts the OpenCode configuration and it is now isolated in #337. I confirmed the **net diff against `main` contains no `opencode.jsonc` or `.gitignore` change** — 11 files, all of them polling code, tests, or the corresponding docs. Scope is clean.

The `Readme.cs` comment was also corrected to say the integration test bounds the *number of retries*, not elapsed time. That is a small thing, but it is accurate now, and the earlier wording would have misled the next reader.

### Residual, not blocking

15 hours is finite but still a long time to block a thread with no way to cancel, and it is reachable from a server that repeatedly sends `Retry-After: 900` or higher. That is an unusual server, the behavior is documented in the changelog, and the real fix is the `CancellationToken` support already listed as an unimplemented gap in AGENTS.md. Out of scope here; worth carrying into whichever PR adds cancellation, so the elapsed-time bound is revisited rather than inherited by default.

### Verdict

**Approve.** All four original findings and the follow-up regression are resolved, the fix is verified rather than asserted, and the unrelated commit is properly split out.

---

**Verification:** Full `net10.0` unit run (152 passed); re-run of the internal-seam probe at `Retry-After` values of 900 and `int.MaxValue` confirming a 15-minute clamp and a 15-hour worst case; `git diff origin/main...848ed8b` confirming the revert leaves no OpenCode configuration in the net diff; review of the new cap assertion; confirmation of PR #337 carrying the split-out config; five green checks on the head commit. Not verified: behavior against a real CA emitting `Retry-After` (the Pebble harness disables validation delays), and `net462` runtime behavior.

fszlin:
The PR description has been updated to match the final implementation and verification. The OpenCode configuration change is confirmed reverted here and isolated in PR #337.

The previous progress comments did not include the required AI disclosure. This comment corrects that for the PR update:

**AI assistance:** Posted with OpenCode using gpt-5.6-luna. AI assisted with PR maintenance and this review-status update. No human review is claimed by this disclosure.

Current status: the latest review approved the implementation, all five hosted checks passed, and the final code includes the 15-minute per-delay `Retry-After` safety cap, explicit retry-budget preservation, stale-date handling, and regression coverage.

fszlin:
Approved to merge per the repository owner’s explicit direction. The owner accepts the documented 15-minute per-delay cap as an intentional policy, including polling earlier than a longer server Retry-After interval. This resolves the outstanding review objection by an explicit design decision and supersedes my previous changes-recommended verdict.

Explicit retry budgets remain preserved, the new default and caller-recompilation implications are documented, and unrelated OpenCode configuration changes have been reverted. Remaining documentation/test follow-ups are non-blocking.

Verification at 848ed8b: 152 local net10.0 unit tests passed, git diff --check passed, and all five hosted checks are now green (Linux, Windows, macOS, package smoke, and Pebble integration). Public-CA interoperability was not verified.

AI-assisted review: OpenCode using gpt-6-astra. AI performed the review and verification; the repository owner explicitly accepted the cap and authorized merging. This comment records that authorization, not an independent human code review.

PR #338, opened 2026-09-21
Fix certificate download with missing alternate links

fszlin:
## Summary
- avoid dereferencing null response links when a preferred chain does not match
- preserve the existing fallback to the default certificate chain when no alternate links are available
- add an offline regression test covering a preferred-chain mismatch with `Links = null`

## Verification
- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true --filter FullyQualifiedName~OrderContextTests`: 2 passed
- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true`: 150 passed, 0 failed, 0 skipped
- `git diff --check`: passed

**AI assistance:** Prepared with OpenCode using gpt-5.6-luna. AI assisted with implementation and PR preparation. No human review is claimed by this disclosure.


fszlin (commented):
## Review of 152d84d

**The implementation looks correct; no blocking code defect found. Merge readiness is blocked by a conflict with current main.**

The explicit null check short-circuits before Links.Contains and preserves the default-chain fallback already used when no alternate relation exists. Existing alternate selection remains unchanged. The new offline regression exercises a nonmatching preferred issuer with null Links, asserts the returned certificate, and uses strict HTTP mocks so unexpected alternate requests fail. It would hit the original null dereference without the fix.

### Before merge
- Update this branch from current main and resolve the conflict in test/Certes.Tests/Acme/OrderContextTests.cs by retaining BOTH FinalizePropagatesRetryAfter from #336 and DownloadFallsBackToDefaultChainWhenAlternateLinksAreMissing from this PR. Preserve the merged production RetryAfter changes as well.
- Re-run verification on the combined branch. The 150 tests here are the older 149-test baseline plus this regression, not the current main baseline plus this regression; expect 153 if no other tests change. GitHub reports CONFLICTING/DIRTY and no hosted checks for this head.

### Minor documentation follow-up
AGENTS.md still lists this exact missing-Link null dereference as an outstanding issue. Remove or revise that entry as instructed by the guide, and add a short changelog fix entry for the user-visible behavior.

### Verification
Reviewed the complete two-file diff and surrounding chain matching/download logic. Re-ran the full offline suite on this exact head: **150 passed, 0 failed, 0 skipped**, macOS ARM64, SDK 10.0.301 / runtime 10.0.9. git diff --check passed against the branch merge base. Existing SourceLink submodule warnings remain. No merged-result tests, local Pebble run, or public-CA interoperability verification was performed. No source edits or branch switches were made; untracked work was preserved.

**AI-assisted review:** OpenCode using gpt-6-astra; review of another session’s implementation, not an independent human review.

fszlin (commented):
## Review: certificate download with missing alternate links

The one-line fix is correct and the regression test is real. Two things block a clean merge, and one closely related bug is left in place.

### The fix and its test hold up

`resp.Links == null ||` is correctly placed before `!resp.Links.Contains("alternate")`, so the short circuit does the work. The loop below is unaffected because `alternateLinks` is materialised with `ToList()` before `resp` is reassigned.

I mutation-tested the regression test by reverting the guard and running it alone:

```
System.NullReferenceException : Object reference not set to an instance of an object.
Failed!  - Failed: 1, Passed: 0
```

So the test genuinely pins the bug rather than passing incidentally. The `Times.Once` verification on the certificate POST is a good touch — it proves no alternate fetch was attempted, not merely that the call returned something.

### 1. The branch conflicts with `main`, and no CI has run

`gh` reports `mergeable: CONFLICTING` and an **empty check rollup — no hosted run has executed on this PR at all**. #336 merged at 03:18 and touched both `src/Certes/Acme/OrderContext.cs` and `test/Certes.Tests/Acme/OrderContextTests.cs`; this branch predates it, so the reported "150 passed" was measured against a stale base.

I resolved it locally to check the fix survives. Only `OrderContextTests.cs` actually conflicts — both PRs appended a test method at the same point — while `OrderContext.cs` auto-merges cleanly, because #336 touched `Finalize` and this touches `Download`. On the merged result:

```
Passed!  - Failed: 0, Passed: 153, Skipped: 0
```

So the fix is compatible; the branch just needs updating and a green hosted run before merge. Worth stating plainly: local numbers against a superseded base are not evidence about what will merge.

### 2. The same bug is still present in `OrderListContext`

`src/Certes/Acme/OrderListContext.cs:42` has the identical pattern with no guard:

```csharp
next = resp.Links["next"].FirstOrDefault();
```

I confirmed it with a probe:

```
PROBE OrderListContext.Orders() -> NullReferenceException
```

Per RFC 8555 §7.1.2.1 the `next` link is present only while more pages remain, so its **absence is the normal loop-termination path**. Any server that returns a complete order list without a `Link` header makes `Orders()` throw immediately. That is arguably a worse instance than the one being fixed, and it is the only other `.Links` dereference in `src/`. Given the PR title is the general "missing alternate links", I would fix both here.

### 3. A root fix would close the class instead of the instances

`ExtractLinksFromResponse` returns `default(ILookup<string, Uri>)` — that is, `null` — when no `Link` header is present. Returning an **empty lookup** instead removes the whole category at a single site, because `ILookup`'s indexer returns an empty sequence for a missing key rather than throwing. Both call sites then work with no guards at all: `Contains("alternate")` is `false`, and `Links["next"].FirstOrDefault()` is `null`, which terminates the loop correctly.

`AcmeHttpResponse.Links` is public, but moving from sometimes-null to never-null is safe for consumers that null-check today. Your call whether to do that here or keep this PR minimal — but if it stays minimal, item 2 should still be covered.

### 4. Stale baseline entry

`AGENTS.md:244` still lists this as a known defect:

> `OrderContext.Download()` can dereference null `Links` when a preferred chain does not match…

#336 updated its baseline entries when it fixed them; this PR should do the same so the list keeps meaning something.

### Minor

The disclosure names `gpt-5.6-luna` concretely, which is what AGENTS.md asks for and an improvement over #336's "the current session model".

### Verdict

**Approve the change itself; do not merge yet.** Rebase onto `main` and get an actual green run, and please either fix `OrderListContext` here or file it — proving the identical crash one file away and shipping only half of it is the outcome worth avoiding.

---

**Verification:** Full `net10.0` unit run on the branch (150 passed); mutation test reverting the guard to confirm the new test fails with `NullReferenceException`; local conflict resolution against current `main` followed by a full run of the merged result (153 passed); a probe demonstrating `OrderListContext.Orders()` throwing `NullReferenceException` with no `Link` header; `git grep` confirming those are the only two `.Links` dereferences in `src/`; inspection of PR mergeability and the empty check rollup. Not verified: behavior against a real CA that omits the header, and `net462` runtime behavior.

fszlin:
Updated PR #338 from current `main` and addressed the review feedback.

Changes:
- Rebases the branch onto current `origin/main`, retaining both the merged `FinalizePropagatesRetryAfter` test and the missing-alternate-links regression.
- Preserves the default-chain fallback when `OrderContext.Download()` receives no `Link` header.
- Fixes the related normal pagination termination case in `OrderListContext.Orders()` when the `next` link is absent.
- Adds an offline regression test for missing `next` links.
- Removes the resolved stale issue from `AGENTS.md` and documents both fallback behaviors in the changelog.

Verification:
- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true`: 154 passed, 0 failed, 0 skipped
- `git diff origin/main...HEAD --check`: passed
- PR diff is limited to the two production fixes, two focused tests, and related documentation.

**AI assistance:** Posted with OpenCode using gpt-5.6-luna. AI assisted with implementation, review-response work, and PR maintenance. No human review is claimed by this disclosure.

fszlin (commented):
## Re-review of `5fc4242`

All four items resolved. Clean rebase, both crash sites fixed, both covered by tests I confirmed are real.

| # | Item | Status |
| --- | --- | --- |
| 1 | Conflicting branch, no CI | Rebased, `MERGEABLE`, five checks green |
| 2 | Same bug in `OrderListContext` | Fixed with `Links?["next"]`, plus a new test |
| 3 | Root fix via empty lookup | Not taken — acceptable, see below |
| 4 | Stale `AGENTS.md` entry | Removed, changelog entry added |

### Verification

Clean worktree at `5fc4242`: **154 tests pass** on `net10.0`. All five checks are green on the head commit, and the PR is `MERGEABLE`. The branch is now a single commit on current `main`, so what I tested is what will merge.

I mutation-tested both fixes by reverting each guard and running its test alone:

```
Links?["next"]        reverted -> System.NullReferenceException, Failed: 1
Links == null || ...  reverted -> System.NullReferenceException, Failed: 1
```

Both tests fail for the right reason when their fix is removed, so neither passes incidentally.

`Links?["next"].FirstOrDefault()` is the right shape for the pagination case: a null lookup yields null, which terminates the `while (next != null)` loop exactly as an absent `next` link should. The new `OrderListContextTests` uses `MockBehavior.Strict` and asserts `Times.Once` on the page fetch, so it would catch a spurious second request rather than only the absence of a crash — that is a stronger assertion than the bug strictly required.

`git grep` confirms all three `.Links` dereferences in `src/` are now guarded or safely sequenced.

### On the root fix

Keeping the null guards rather than making `ExtractLinksFromResponse` return an empty lookup was explicitly your call, and the minimal version is defensible for a bugfix PR — it avoids changing a public property's nullability alongside two behavior fixes.

Worth recording that the underlying shape is unchanged: `Links` is still sometimes-null, so a future `.Links` use starts out crashing by default rather than working by default. Both current sites are now correct, so this is a note for whoever adds the third one, not an objection to this PR.

### Verdict

**Approve.** Ready to merge.

---

**Verification:** Full `net10.0` unit run at `5fc4242` (154 passed); separate mutation tests reverting each of the two guards, both producing `NullReferenceException` in their own test; `git grep` over `src/` confirming every `.Links` dereference is guarded; review of the new `OrderListContextTests` for strict-mock and call-count assertions; confirmation of `MERGEABLE` and five green checks on the head commit. Not verified: behavior against a real CA that omits the headers, and `net462` runtime behavior.

PR #339, opened 2026-09-21
Export certificate chains without requiring a self-signed root

### LINKED ISSUE #315: Planned changes to Lets Encrypt - short chain delivery
On the 9th November Lets Encrypt released a changes to production by accident that caused an error in Certes.

**Error Message:** Can not find issuer 'C=US,O=Internet Security Research Group,CN=ISRG Root X1' for certificate 'C=US,O=Let's Encrypt,CN=R3'

They investigated and put there hands up and said a change had crept in to the release that was not meant to happen until Feb 2024.  So the issue that caused the error will be released onto live server very soon.

This change involves not returning the root certificate with the request (short chain).  You will need to excuse me but I'm no expert in the field but here is a link to the post explaining what happened.

https://community.letsencrypt.org/t/can-not-find-issuer-c-us-o-internet-security-research-group-cn-isrg-root-x1-for-certificate-c-us-o-lets-encrypt-cn-r3/208268/1

Here is the announcement by Lets Encrypt:   https://letsencrypt.org/2023/07/10/cross-sign-expiration.html

Is there a plan to update this library to work with this change?

Thanks

Rob

fszlin:
## Description

Both PEM and PFX export walked the issuer chain until they reached a self-signed root and threw `AcmeException` otherwise. RFC 8555 section 7.4.2 says the ACME server **SHOULD NOT** include that root in the returned chain, so the spec-compliant response was exactly the input that failed. This worked in practice only because three Let's Encrypt roots are embedded in the library; against Pebble, ZeroSSL, Buypass, or any internal CA, export threw on a valid chain.

Likely resolves #315, which is the user-visible form of this bug: Let's Encrypt's short-chain delivery produced `Can not find issuer ... for certificate ...`. See [the comment there](https://github.com/fszlin/certes/issues/315#issuecomment-5760729655) for what this PR does and does not explain — the original embedded-root lookup failure is still undiagnosed, but is no longer fatal.

### Changes

- `CertificateStore.GetIssuers()` stops at the highest issuer available rather than requiring a self-signed root.
- Issuers are accepted only after verifying that they signed the certificate below them. Candidates are indexed by subject name, which does not establish an issuer relationship, so a certificate sharing an issuer's name but holding a different key was previously accepted — PEM emitted it into the exported chain, and PFX silently dropped it and packaged the leaf alone.
- `PfxBuilder.FindIssuers()` falls back to the linked issuers when no trust anchor is present, because `PkixCertPathBuilder.Build` fails on an empty anchor set.
- The issuer walk is guarded against cycles, returning the verified acyclic prefix. Subject-name indexing allows two mutually cross-signed certificates to form a loop.

### PFX output is unchanged

The PFX path already excluded the self-signed root and used it only as a trust anchor for path building — it was refusing to export over a certificate it would then discard. `PfxChainExcludesRootWhenSupplied` pins that shape so the claim is enforced rather than asserted.

### Behavior changes

Two, both deliberate:

1. A chain reaching an intermediate but no root now exports instead of throwing. This is the point of the PR.
2. Supplying no issuers at all now exports the certificate alone, rather than throwing. A leaf issued directly by a root the server omitted is legitimate and was previously indistinguishable from a broken chain. This renamed one existing test, `FullChainRequiresIssuers` → `FullChainExportsLeafWhenNoIssuersAreSupplied`.

Supplying issuers that did not sign the certificate still fails; `FailWhenMissingIntermediateCert` passes unmodified.

### Out of scope

Embedded roots remain an automatic fallback for both formats, and subject-name indexing is unchanged — so cross-signed alternates still cannot coexist in the store, and signature verification can reject a wrong-key candidate but not select among same-subject alternates. Issuer constraints (basic constraints, key usage, validity) are not enforced; packaging is not path validation.

Tracked as #340 (prerequisite, and the reason no alternate-path-selection tests are included here) and #341. `AGENTS.md` records these limitations in place of the stale "requires a path to a self-signed root" entry.

## Verification

SDK 10.0.301, macOS ARM64.

- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true` — **167 passed, 0 failed, 0 skipped** (154 on `main`; 13 net new, 1 renamed).
- `dotnet build src/Certes/Certes.csproj -c Release` — all library targets.
- `dotnet build test/Certes.Tests/Certes.Tests.csproj -f net462 -p:SkipSigning=true` — compiles; not run.
- `git diff --check` — clean. All five hosted checks pass.

Each change was mutation-tested independently:

| Mutation | Result |
| --- | --- |
| Restore the root requirement | 4 tests fail |
| Remove `certificate.Verify(issuer.GetPublicKey())` | 3 tests fail |
| Disable the PFX anchor fallback | 2 fail — independently load-bearing, not masked by the store fix |
| Remove the cycle guard | Test does not terminate within 90s |

**Not verified:** behavior against a real CA, `net462` at runtime, and whether Azure App Service accepts a rootless PFX. That last point is relevant to `AzureAppCommand` users, though that path already produced rootless PFXs before this change.

## Checklist

- [x] All tests are passing
- [x] Documentation updated: changelog entry and `AGENTS.md` baseline

## AI assistance

**Review status:** Reviewed across two rounds by a separate OpenCode session (gpt-6-astra, then gpt-5.6-luna), which found the wrong-key issuer defect and the immediate-issuer policy problem — both fixed in `f5e0391`. One finding was disputed rather than applied: the follow-up review's P1 held that the cycle guard returns an invalid chain, and I showed the returned prefix is acyclic with every link signature-verified, so only the test assertion was strengthened. That disagreement is unresolved, not conceded. No human review is claimed.


fszlin (commented):
## Review of 45ba29b

**Changes recommended before merge: the new rootless PFX path accepts an invalid issuer relationship and silently exports an incomplete key chain.**

### [P1] Verify issuer links before bypassing PKIX — src/Certes/Pkcs/PfxBuilder.cs:118–126

The fallback treats CertificateStore output as linked issuers, but that store selects by subject DN only; it never verifies the child signature. Returning the list directly removes the validation that FindIssuers previously performed. An unrelated CA certificate with the expected subject name is therefore accepted as an intermediate.

I reproduced this using two independently generated CertificateFixture instances (their issuer DNs match but their keys differ):

```csharp
var a = new CertificateFixture(KeyAlgorithm.ES256);
var b = new CertificateFixture(KeyAlgorithm.ES256);
// a.Leaf.Verify(b.Intermediate.GetPublicKey()) throws: not its issuer.
var chain = CertificateFixture.ChainOf(a.Leaf, b.Intermediate);
var pfx = chain.ToPfx(a.Key).Build("review", "test-password");
```

Build unexpectedly succeeds. Reloading it with Pkcs12Store and calling GetCertificateChain for the key alias returns **one certificate (the leaf)**, not leaf + intermediate. I verified both results with temporary diagnostic tests: the expected rejection assertion failed because no exception was thrown, and the expected chain-length assertion reported expected 2 / actual 1. This is not a request to restore root trust validation: verify child/issuer signatures and suitable issuer constraints independently of whether a trust anchor exists, reject invalid supplied relationships, and retain exact-certificate/key-association regression tests. Same-subject/different-key candidates must not be treated as equivalent issuers. This is observable on the newly enabled path even though subject-name indexing itself predates the PR.

### Policy/scope gaps to acknowledge

The implementation is narrower than the export policy I recommended in our discussion:
- `CertificationStore.cs:72–77` still throws when no immediate issuer is available. That includes a valid leaf signed directly by an omitted root, as well as intentional leaf-only packaging; chain.Count == 0 does not establish that supplied issuers are wrong. The existing negative tests encode the old policy, not a general certificate validity rule. If this is intentionally deferred, document that limitation explicitly rather than claiming missing roots are generally supported.
- PFX still uses embedded roots automatically through the shared store, and still runs PKIX validation when one is found. My recommendation was to retain that fallback for PEM compatibility only and separate PFX packaging from trust validation. The PR does explicitly defer this, so please confirm the narrower scope rather than treating that recommendation as implemented.
- The new subject-DN cycle guard terminates traversal, which is useful, but it is not coverage for distinct same-subject/cross-signed candidates. Current tests cover a single happy-path rootless chain and an A/B cycle, not alternate path selection or private-key mismatch handling.

### Minor fixes
- Update AGENTS.md: it still says chain export requires a path to a self-signed root. Describe the remaining immediate-issuer and embedded-root limitations accurately.
- Restore the Allman brace/newline accidentally collapsed on CertificateFixture.AssertPfx (line 87).

### Verification
- Full unmodified PR suite: **163 passed, 0 failed, 0 skipped** on macOS ARM64, SDK 10.0.301 / runtime 10.0.9.
- Release core-library build passed for net10.0, net8.0, netstandard2.0 and net462.
- git diff --check passed.
- Two additional temporary diagnostic assertions failed as described above; removed after reproduction. Working tree is clean.
- Hosted Linux, macOS, package smoke and Pebble checks passed; Windows was pending at the last observation.
- Known SourceLink submodule warnings remain. No local Pebble, Framework-runtime, public-CA, or Azure App Service verification performed.

**AI-assisted review:** OpenCode using gpt-6-astra. Review of another session’s implementation with independent diagnostic tests; not an independent human review.

fszlin:
Addressed in `f5e0391`.

## [P1] Valid, and it was worse than the reproduction showed

Confirmed and fixed. The fallback returned subject-name matches without the signature checking PKIX had been providing implicitly.

One correction to the report: PFX is the *milder* of the two. Your reproduction found `GetCertificateChain` returning one certificate, because `Pkcs12Store` refuses to link a certificate it cannot chain — so the bad issuer was dropped. **PEM had no such backstop and emitted the unrelated intermediate into the exported chain.** That makes a bogus `fullchain.pem` served to clients, which is a worse outcome than a truncated PFX.

So I fixed it in `CertificateStore.GetIssuers()` rather than in `PfxBuilder`, which covers both formats at the point where the wrong decision is actually made. A candidate is now accepted only if `certificate.Verify(issuer.GetPublicKey())` succeeds.

Your exact reproduction now yields:

```
pfx=AcmeException  pem=AcmeException
"Can not find issuer 'CN=Certes Unit Test Intermediate' for certificate 'CN=unit.example'."
```

Covered by three tests — store level, `ToPem`, and `ToPfx`. Removing the `Verify` call fails exactly those three and nothing else.

This does not implement issuer *constraint* checking (basicConstraints, keyUsage, validity) — packaging still is not path validation. Same-subject/different-key candidates are now rejected, which was the concrete defect.

## Immediate-issuer policy: you were right, and it had a better answer

You flagged that `chain.Count == 0` does not establish that supplied issuers are wrong, and that it wrongly rejects a leaf signed directly by an omitted root. Agreed — but rather than documenting it as a deferred limitation, the signature check makes the distinction expressible, so I fixed it:

- **Issuers supplied, none signed the certificate** → still throws. `FailWhenMissingIntermediateCert` passes unmodified.
- **No issuers supplied at all** → exports the certificate alone. A leaf issued directly by an omitted root now works.

This changes one existing test: `FullChainRequiresIssuers` became `FullChainExportsLeafWhenNoIssuersAreSupplied`. That is a deliberate behavior change and the single test I modified — calling it out rather than leaving it in the diff.

## Scope, confirmed rather than implied

Confirming the narrower scope explicitly, as asked. **Not** implemented here:

- Embedded roots remain an automatic fallback for both PEM and PFX; PFX still runs PKIX validation when a root is found. Splitting packaging from trust validation, and the PEM-compatibility-only policy, are not in this PR.
- Subject-DN indexing is unchanged. The store keeps one candidate per subject name, so cross-signed alternate selection is still impossible — the cycle guard bounds traversal and is not a substitute. Signature verification now rejects a wrong-key candidate, but cannot *choose* the right one among same-subject alternates, since only one is retained.

Both need the indexing change first and belong in their own PRs. `AGENTS.md` now states these limitations instead of the stale "requires a path to a self-signed root" claim, and the changelog no longer implies missing roots are generally supported.

## Minor

- `AGENTS.md` updated as described.
- Allman brace on `AssertPfx` restored — that was my editing error, thanks for catching it.

## Verification

SDK 10.0.301, macOS ARM64.

- Full suite: **167 passed, 0 failed, 0 skipped** (163 before; 4 net new, 1 renamed).
- `dotnet build src/Certes/Certes.csproj -c Release` — all targets. `net462` test project compiles; not run.
- `git diff --check` — clean.
- Mutation tests: removing `Verify` fails the 3 new tests; restoring the root requirement fails 4; disabling the PFX anchor fallback fails 2; removing the cycle guard does not terminate within 90s.

Unchanged from before: no local Pebble, Framework-runtime, public-CA, or Azure App Service verification. The Azure question is worth a second opinion — that path already produced rootless PFXs before this PR, so I do not believe this changes it, but I have not tested against App Service.

**AI assistance:** Response prepared with OpenCode using claude-opus-5 — the same session that authored the PR, so this is a self-response to your review, not independent verification of it.

fszlin:
## Follow-up Review Findings

The latest commit addresses the wrong-key issuer defect and the no-supplied-issuer policy, but two compatibility risks remain:

1. **[P1] Cycle handling still returns an invalid chain**
   - `src/Certes/Pkcs/CertificationStore.cs:87-98`
   - The visited-set guard prevents infinite traversal but returns the accumulated cyclic certificates as a successful issuer chain.
   - `GetIssuersTerminatesOnMutuallyIssuedCertificates` explicitly expects the two-certificate cycle to be returned.
   - The cycle can therefore still be exported through PEM and can enter the rootless PFX fallback.
   - Recommendation: reject cycles, or return only a clearly defined validated acyclic prefix. The regression should assert the intended failure/partial behavior rather than `Count == 2`.

2. **[P2] Cross-signed/duplicate-subject selection remains unresolved**
   - `src/Certes/Pkcs/CertificationStore.cs:16,45,108-109`
   - Signature verification now rejects wrong-key candidates, which fixes the concrete defect previously identified.
   - However, the dictionary still retains only one certificate per `SubjectDN`, so a valid alternate/cross-signed issuer can be overwritten before verification.
   - No alternate/cross-signed path-selection test has been added.

3. **[P2] Rootless PFX fallback bypasses full path validation**
   - `src/Certes/Pkcs/PfxBuilder.cs:118-126`
   - The fallback verifies issuer signatures through `CertificateStore`, but does not run full PKIX validation when no self-signed root is available.
   - Constraints such as basic constraints, key usage, validity, path ordering, and cycles are not fully validated.
   - This is documented as intentional scope, but remains a compatibility/security limitation.

### Addressed

- Wrong-key/same-subject issuer acceptance: fixed with `certificate.Verify(issuer.GetPublicKey())`.
- Leaf-only export when no issuer material is supplied: fixed and covered.
- Missing-root PEM/PFX export: fixed and covered.
- Supplied issuers that do not sign the leaf: still rejected.

All five hosted checks pass. No source changes were made in this review.

**AI-assisted review:** Posted with OpenCode using gpt-5.6-luna. AI assisted with this review and PR comment. This is not an independent human review, and no human approval is claimed.

fszlin:
## [P1] I disagree — the returned chain is neither cyclic nor unverified

The finding says the guard "returns the accumulated cyclic certificates as a successful issuer chain" and that this can be exported. I checked that directly rather than reasoning from the test name, by re-verifying every link in the returned output:

```
count=2  acyclic=True
[CN=cycle.example     <- CN=Certes Cycle A: VERIFIED]
[CN=Certes Cycle A    <- CN=Certes Cycle B: VERIFIED]
```

No subject name repeats, and **every link is signature-verified** — `leaf` really was signed by A, and A really was signed by B. The guard stops *before* re-adding A, so the cycle is never entered into the result.

Your recommendation was to "return only a clearly defined validated acyclic prefix". That is what the code already returns. The output is not a cycle; it is a verified prefix that happens not to terminate at a self-signed root, which is the same truncation case this PR deliberately permits.

On exportability: a PEM or PFX containing `leaf → A → B` creates no false trust. The relying party builds the path against its own trust store, and since neither A nor B is anchored there, validation fails exactly as it would for any chain that does not reach a trusted root. Rejecting it outright would also reject legitimate truncation, because the two are indistinguishable at this layer — the only thing that makes this input a "cycle" is material the store never sees.

**What was fair in the finding:** `Assert.Equal(2, issuers.Count)` was a weak assertion that did not express *why* two is correct, which is presumably what made the behavior look like endorsement of cycle output. Fixed in `9087ed8` — the test is now `GetIssuersReturnsVerifiedAcyclicPrefixForMutuallyIssuedCertificates` and asserts the invariant: no subject name is revisited, and every returned issuer signed the certificate below it. The behavior is unchanged; the test now states the contract.

If you still want cycles rejected, that needs an argument for what concrete harm the verified prefix causes, because it would cost the legitimate truncation case.

## [P2] Cross-signed selection, [P2] PFX path validation

Both confirmed as real and deliberately outside this PR — your summary matches mine, and I am not claiming either is fixed.

Both depend on the same prerequisite: `Dictionary<X509Name, X509Certificate>` retains one candidate per subject name, so a valid cross-signed alternate can be overwritten at `Add` time, before verification ever runs. Signature verification rejects a wrong-key candidate but cannot select among same-subject alternates when only one survives insertion. Alternate-path-selection tests are not meaningful until the store can hold alternates, which is why none were added here rather than an oversight.

Sequencing I would suggest, one PR each:

1. Index issuers by subject name **and** subject key identifier, so alternates coexist. Adds cross-signed selection tests.
2. Separate PFX packaging from trust validation, and decide the embedded-roots policy — including removing the expired DST Root CA X3 and the Fake LE **staging** root from automatic fallback.

Item 2 is where issuer constraints — basic constraints, key usage, validity, path ordering — belong, since that is path validation rather than packaging. `AGENTS.md` now records all of this as current limitations.

Happy to open tracking issues for both if you want them on the backlog rather than in review comments.

## Verification

167 passed, 0 failed, 0 skipped on `net10.0`; `git diff --check` clean. No production code changed in `9087ed8` — test assertions only.

fszlin:
Context and follow-ups:

- Likely resolves #315 (Let's Encrypt short-chain delivery). That issue is the user-visible form of this bug: export threw `AcmeException` on a chain with no root. I have commented there on what this PR does and does not explain — the original embedded-root lookup failure is still undiagnosed, but is no longer fatal.
- Follow-ups filed as agreed, both explicitly out of scope here: #340 (cross-signed alternates cannot coexist in `CertificateStore`) and #341 (remove automatic embedded-root fallback; separate PFX packaging from path validation).

#340 is the prerequisite for the alternate-path-selection tests raised in review.

fszlin (commented):
## Readiness re-review of 9087ed8

**Ready to merge for the explicitly narrowed packaging scope. My original wrong-key issuer blocker is resolved.**

I read the review discussion, checked the latest diff, and reran the full unit suite rather than relying solely on the response comments.

- The store verifies each candidate signature before including it. The new store/PEM/PFX regressions cover my same-subject/wrong-key reproduction and pass.
- No-supplied-issuer export now succeeds, including leaf-only PFX, with certificate/key and exact-chain assertions.
- AGENTS.md and the changelog describe the new behavior and remaining embedded-root/path-validation limitations.

### On the disputed cycle finding
The input issuer graph can contain a cycle, but the returned ordered list stops before repeating a subject. Each included child/issuer link is signature-verified. Under this PR’s stated contract of packaging an available prefix rather than proving a trusted path, I do not consider that a merge blocker. The revised invariant test makes the contract clearer. This does not establish trust or turn the cyclic issuer graph into a valid complete certification path; consumers still need their own validation.

### Deferred scope
Duplicate-subject alternate selection and consistent packaging-versus-validation behavior remain real limitations, but are explicitly acknowledged and tracked in #340 and #341. Full issuer-constraint validation is not claimed by this review. Those broader changes need not block this focused rootless-export fix.

Two non-blocking documentation corrections:
1. Update the PR description: it still says FullChainRequiresIssuers passes unmodified, claims no link always throws, and reports 163 tests. The current behavior and count are different.
2. Correct the proposed identity scheme in #340 before implementation. Cross-signed variants of the SAME certificate key (including self-signed and DST-cross-signed ISRG Root X1) share the public key and typically the SKI. Subject + SKI alone will still collapse those variants, and verifying a child signature cannot distinguish their alternate paths. Retain distinct certificates by DER/fingerprint within the candidate collection, then select a path using issuer/anchor preferences. Different-key same-subject collisions are a separate case.

### Verification
At 9087ed8: **167 passed, 0 failed, 0 skipped** on net10.0, SDK 10.0.301 / runtime 10.0.9, macOS ARM64. git diff --check passed. All five hosted checks passed and GitHub reports CLEAN merge status. Known SourceLink warnings remain. I did not repeat the all-target Release build from my earlier review, or locally verify Pebble, public CAs, Framework runtime, or Azure App Service. No source changes made.

**AI-assisted review:** OpenCode using gpt-6-astra. Follow-up review of another session’s implementation, not an independent human review. This readiness assessment supersedes my prior changes-recommended verdict; it does not itself authorize merging.

Issue #341, opened 2026-09-21
Remove automatic embedded-root fallback and separate PFX packaging from path validation


fszlin:
## Problem

The library embeds three CA roots and consults them automatically during chain export for both PEM and PFX (`src/Certes/Certes.csproj`, `CertificateStore.embeddedCertificates`):

| Resource | Status |
| --- | --- |
| `dst-root-ca-x3.pem` | **Expired 2021-09-30** |
| `fake-le-root-x1.pem` | **Let's Encrypt staging root**, valid to 2036 |
| `isrg-root-x1.pem` | Current |

Three problems:

1. An expired root can be injected into an exported chain automatically. Serving DST Root CA X3 after its expiry is what broke OpenSSL 1.0.2 clients in September 2021.
2. A **staging** trust anchor ships in the production package and is reachable by silent fallback.
3. Certes is CA-agnostic, but only Let's Encrypt is bundled — and incompletely. ISRG Root X2 is absent, as is every other ACME CA. Roots rotate faster than this library's release cadence.

This was raised upstream in #315 by a Let's Encrypt engineer: *"What changes would be necessary to remove this root bundling entirely? It seems like an easy and error-prone way for clients to get out of date (e.g. it doesn't contain ISRG Root X2, it doesn't include any other ACME CA's roots, and it does contain LE's Staging root...)"*

PR #339 removed the *reason* the bundle existed — export no longer requires reaching a self-signed root — so the automatic fallback now has no remaining function. Callers who want a root packaged can supply it through `AddIssuer`/`AddIssuers`.

## Related: separate PFX packaging from path validation

`PfxBuilder.FindIssuers` runs PKIX validation when a self-signed root is available and packages the linked issuers otherwise, so the amount of validation depends on whether a root happens to be present. Issuer constraints — basic constraints, key usage, validity, path ordering — are not consistently enforced either way.

Packaging and trust validation are different operations and should not be coupled to root availability. If Certes validates, it should do so consistently; if it packages, it should say so.

## Proposal

1. Stop consulting embedded roots automatically. Transitional opt-in flag, or removal in the next major with a changelog entry.
2. Remove `dst-root-ca-x3.pem` and `fake-le-root-x1.pem` from the package regardless of the outcome of (1). Neither belongs in a production artifact.
3. Decide explicitly whether PFX export validates or packages, and apply it consistently.

## Notes

[...]

PR #342, opened 2026-09-21
Retain cross-signed issuer alternates

### LINKED ISSUE #340: CertificateStore cannot hold cross-signed issuer alternates
## Problem

`CertificateStore` indexes issuers by subject name only:

```csharp
private readonly Dictionary<X509Name, X509Certificate> certificates = new Dictionary<X509Name, X509Certificate>();
...
this.certificates[cert.SubjectDN] = cert;   // last write wins
```

Cross-signed certificates legitimately share a subject name while holding different keys — ISRG Root X1 exists both self-signed and cross-signed by DST Root CA X3. When both are supplied, the second silently replaces the first, so:

- only one candidate per subject name can ever be considered, and
- the surviving candidate may be the wrong one for the chain being built.

PR #339 added signature verification, so a wrong-key candidate is now rejected rather than emitted. But rejection is all it can do: if the correct alternate was already overwritten at `Add` time, there is nothing left to select, and export fails instead of choosing the valid path.

This also blocks meaningful alternate-path tests. Cross-signed selection cannot be tested while the store can only hold one alternate, which is why PR #339 added none.

## Proposal

Key candidates by subject name **and** subject key identifier, keeping all alternates, and select during the walk by verifying the signature — which `GetIssuers` already does per candidate. Authority key identifier can narrow the candidate set before verification where present.

## Acceptance

- Both ISRG Root X1 variants can coexist in one store.
- `GetIssuers` selects the alternate that actually signed the certificate below it.
- Coverage for cross-signed alternate selection, including the case where the wrong alternate is added last.
- `preferredChain` selection exercised against a store holding both alternates.

## Notes

Prerequisite for decoupling PFX packaging from trust validation, and for any work on `preferredChain` against cross-signed roots. Baseline recorded in `AGENTS.md`.

fszlin:
Closes #340.

## Description

`CertificateStore` indexed issuers by subject name with last-write-wins semantics:

```csharp
this.certificates[cert.SubjectDN] = cert;
```

Cross-signed certificates legitimately share a subject name — ISRG Root X1 exists both self-signed and cross-signed by DST Root CA X3 — so supplying both kept only whichever was added last. Two consequences:

1. **The exported chain depended on insertion order.** Keeping the cross-signed alternate continues the chain through DST Root CA X3; keeping the self-signed one ends it there. Callers had no way to influence which, and the same inputs in a different order produced different output.
2. **A certificate sharing an issuer's subject name could displace the real issuer** before signature verification had a chance to reject it. #339 added verification, but verification can only reject a candidate — if the correct alternate was already overwritten at `Add` time, there was nothing left to select and export failed.

### Changes

- Retain every distinct alternate per subject name, rather than replacing. Duplicates of the same certificate are ignored.
- Choose the issuer when the chain is built: verify which candidates signed the certificate below them, prefer candidates inside their validity period, then prefer a self-signed alternate so the chain ends there instead of continuing through a cross-signing issuer. Remaining ties keep insertion order.
- Supplied issuers take precedence over the embedded certificates, which are consulted only when nothing supplied can serve as the issuer, preserving the previous embedded-root policy.

The self-signed preference means the **short chain is the default** when both alternates are supplied. That matches how the ecosystem moved after the DST Root CA X3 expiry, and avoids extending a chain through a cross-signing root that may be expired. Supplying only the cross-signed alternate still produces the long chain, so the behavior remains reachable — including when the self-signed variant is one this library embeds.

The validity preference exists because the self-signed preference alone selected an *expired* alternate over a usable cross-signed path, which is the DST Root CA X3 failure shape. Expired candidates remain usable when nothing else verifies, so this changes preference rather than acceptance.

### Behavior change

Where both alternates were supplied, output previously depended on insertion order; it is now deterministic and prefers the short chain. Callers who were implicitly relying on insertion order to obtain the long chain should supply only the cross-signed alternate. Recorded in the changelog.

### Out of scope

The embedded roots remain an automatic fallback for both PEM and PFX, and `PfxBuilder` still runs PKIX validation only when a self-signed root is available. Tracked in #341.

Selection is by signature verification rather than by authority/subject key identifier. AKI/SKI does not disambiguate a true cross-signed pair, because the alternates share a key and therefore a subject key identifier — the only difference is who signed them. AKI would be a useful pre-filter for large candidate sets, not a correctness improvement here.

## Verification

SDK 10.0.301, macOS ARM64.

- `dotnet test test/Certes.Tests/Certes.Tests.csproj -f net10.0 -p:SkipSigning=true` — **175 passed, 0 failed, 0 skipped** (167 on `main`; 8 new).
- `dotnet build src/Certes/Certes.csproj -c Release` — all library targets.
- `dotnet build test/Certes.Tests/Certes.Tests.csproj -f net462 -p:SkipSigning=true` — compiles; not run.
- `git diff --check` — clean.

New coverage uses a synthetic cross-signed PKI mirroring the ISRG Root X1 arrangement: one subject name and key published both self-signed and signed by a second root, plus a same-subject/different-key imposter. Embedded-root precedence is covered against a subject and key the library actually embeds, since synthetic subject names cannot collide with the embedded set.

Mutation-tested:

| Mutation | Result |
| --- | --- |
| Restore last-write-wins in `Add` | 2 fail, including the imposter-shadowing regression |
| Remove the self-signed preference | 1 fails — the insertion-order case the first mutation does not cover |
| Remove the validity preference | 1 fails |
| Rank supplied and embedded candidates together | 1 fails — the embedded-precedence regression |

The two `GetIssuersPrefersSelfSignedCrossSignedAlternate` cases fail under *different* mutations, one each, so order-independence is pinned by both mechanisms rather than incidentally by one.

**Not verified:** behavior against a real CA or against the genuine ISRG Root X1 pair, `net462` at runtime, and `preferredChain` selection end to end against a server offering alternate chains.

## Checklist

- [x] All tests are passing
- [x] Documentation updated: changelog entry and `AGENTS.md` baseline

## AI assistance

**Review status:** Authored by this session, so the checks above are a self-review. No independent or human review is claimed. The short-chain default is the judgment call most worth a second opinion.


fszlin:
## Follow-up Review Findings

1. **[P1] Self-signed preference is not full path selection**
   - `src/Certes/Pkcs/CertificationStore.cs:126-129`
   - Candidates are sorted by `SubjectDN == IssuerDN` after only verifying the immediate child signature.
   - A self-signed same-subject candidate is preferred without checking whether the resulting path is valid, terminates correctly, or satisfies the relevant issuer constraints.
   - Recommendation: choose candidates as part of path construction, preferring a self-signed candidate only when it forms a valid usable path; otherwise continue with a valid cross-signed alternate.

2. **[P1] Cycle handling still returns an invalid chain**
   - `src/Certes/Pkcs/CertificationStore.cs:99-110`
   - The visited-set guard prevents infinite traversal but returns the accumulated cycle as a successful issuer chain.
   - `GetIssuersReturnsVerifiedAcyclicPrefixForMutuallyIssuedCertificates` expects the complete `first, second` cycle, even though `second` points back to `first`.
   - PEM export can emit the cyclic issuer bundle, and the rootless PFX fallback can package the same invalid path.
   - Recommendation: reject repeated subjects/certificates as an invalid path, or return a clearly defined partial chain that excludes the cycle-causing certificate.

3. **[P2] Cross-signed selection is deterministic but not path-validating**
   - Retaining candidates fixes last-write-wins and insertion-order dependence, and the synthetic tests cover the tested topology.
   - The algorithm still uses a local self-signed preference rather than evaluating complete alternate paths.
   - Embedded-root fallback and rootless PKIX limitations remain explicitly out of scope and are documented residual risks.

### Addressed

- Same-subject candidates are retained instead of overwritten.
- Duplicate certificates are ignored.
- Wrong-key imposters are rejected through signature verification.
- Tested self-signed versus cross-signed selection is deterministic.
- Cross-signed-only paths remain supported.

All five hosted checks pass. No source changes were made in this review.

**AI-assisted review:** Posted with OpenCode using gpt-5.6-luna. AI assisted with this review and PR comment. This is not an independent human review, and no human approval is claimed.

fszlin:
## [P1] Self-signed preference — you were right, fixed in `7f0fad9`

I tested the concrete case rather than reasoning about it, and it reproduces:

```
expired self-signed alternate (notAfter 2025-09-16), valid cross-signed path also available
chain: CN=Certes Shared Issuer(exp=2025-09-16)
```

Preferring self-signed on subject name alone selected the **expired** alternate and truncated the chain there, discarding a usable cross-signed path to a valid root. That is the DST Root CA X3 shape — serving an expired certificate in the chain is exactly what broke older clients in 2021. Good catch; the local preference was doing real harm, not just failing to be thorough.

Candidates inside their validity period are now ordered ahead of expired ones, before the self-signed preference applies. Expired candidates remain *usable* when nothing else verifies, so this changes preference rather than acceptance — an expired-only chain still exports rather than newly failing.

Covered by `GetIssuersPrefersValidAlternateOverExpiredSelfSignedOne`; removing the validity ordering fails exactly that test.

**What I did not implement:** selection driven by complete path construction. Validity is one constraint, chosen because it maps to the demonstrated failure. Basic constraints, key usage, path length and full path validity are still not evaluated, and I have not tried to make this a path builder — that is the "separate packaging from trust validation" work in #341, where it belongs. The XML docs and `AGENTS.md` now say so explicitly rather than implying the selection is validating.

## [P1] Cycle handling — still disagreeing, and the evidence is unchanged

This is the same finding from #339, re-raised with the same wording. My refutation there was empirical, so repeating the claim does not move it. Re-running against this branch:

```
count=2  acyclic=True
[CN=cycle.example  <- CN=Certes Cycle A: VERIFIED]
[CN=Certes Cycle A <- CN=Certes Cycle B: VERIFIED]
```

The finding says the guard "returns the accumulated cycle". It does not. No subject name appears twice in the output, and every link is signature-verified: the leaf really was signed by A, and A really was signed by B. The guard stops *before* re-adding A. Your recommendation — "return a clearly defined partial chain that excludes the cycle-causing certificate" — describes what the code already returns.

"`second` points back to `first`" is true and is not a property of the returned chain. It means the chain does not terminate at a self-signed root, which is the truncation case this library deliberately supports since #339, and which is indistinguishable at this layer from any other short chain.

On the stated consequence — that PEM and PFX can emit this — a bundle of `leaf → A → B` creates **no false trust**. The relying party builds the path against its own trust store; neither A nor B is anchored there, so validation fails exactly as it would for any chain not reaching a trusted root. Rejecting it would also reject legitimate truncation.

I will change this if you can show a case where the emitted prefix causes a verifier to **incorrectly accept** a certificate, rather than to fail. That is the standard I am applying to P1 above, which met it — the expired-alternate case produced a demonstrably worse chain, so I fixed it. This one has not.

## [P2] Deterministic but not path-validating

Agreed and unchanged. With `7f0fad9` the selection is deterministic, signature-verified, and validity-aware, and still is not path validation. Recorded in `AGENTS.md` as a limitation, tracked in #341.

## Verification

173 passed, 0 failed, 0 skipped on `net10.0`; Release build of all library targets; `net462` compiles; `git diff --check` clean.

Mutation coverage now:

| Mutation | Result |
| --- | --- |
| Restore last-write-wins in `Add` | 2 fail |
| Remove the self-signed preference | 1 fails |
| Remove the validity preference | 1 fails |

fszlin (commented):
## Review of 7f0fad9

**Changes recommended before merge: embedded roots now override explicitly supplied cross-signed issuers.** Retaining distinct certificates and checking signatures fixes last-write-wins, and the validity preference addresses the demonstrated expired-alternate case. However, candidate-source precedence changed unintentionally relative to the documented scope.

### [P1] Preserve supplied-chain precedence over embedded fallback
`src/Certes/Pkcs/CertificationStore.cs:133–137,142–157`

GetCandidates yields supplied AND embedded certificates, then the global sort ranks them together. A valid embedded self-signed certificate outranks a valid supplied cross-signed variant of the same key. Previously the embedded lookup occurred only when the supplied subject lookup failed. This is a change to embedded-root policy, despite that policy being described as untouched/out of scope, and contradicts the PR statement that supplying only the cross-signed alternate preserves the long chain.

I reproduced this offline without any CA private key:
1. Load the existing fake-le-intermediate-x1.pem fixture and the embedded Fake LE Root X1 public certificate.
2. Generate an ephemeral root, and issue a cross-signed certificate containing the embedded root public key and subject, signed by that ephemeral root.
3. Verify the intermediate signature with the cross-signed public key, and the cross-signature with the ephemeral root.
4. Supply ONLY that cross-signed issuer and its ephemeral root to CertificateStore.
5. GetIssuers(intermediate DER) returns the embedded self-signed Fake LE Root X1 instead of the explicitly supplied cross-signed path. The temporary regression asserting that the embedded root is absent fails.

This also affects export after an alternate chain was deliberately selected: ToPem/ToPfx rebuild through this store, so the embedded self-signed variant can replace the chosen cross-sign. For example, a supplied ISRG Root X1 cross-sign competes with the built-in ISRG Root X1. The caller cannot obtain the documented long-chain behavior simply by omitting the self-signed certificate from Add.

Apply validity/self-signed ranking within usable SUPPLIED candidates first, and consult embedded candidates only as fallback when no supplied candidate can serve as issuer, or make any override an explicit policy change with an opt-out. Add a regression involving an actual embedded subject/key as above; synthetic-only subject names cannot expose this interaction. Assert exported chain contents as well as the store result.

### [P2] Narrow the changelog claim about insertion order
`docs/CHANGELOG.md:15–17` says exported chains no longer depend on insertion order, but TryGetIssuer explicitly retains insertion order for equal-ranked candidates. Two currently valid cross-signs of the same key under different roots still select whichever was added first, and can lead to different chains. The XML docs and PR description acknowledge ties; make the changelog equally precise (order-independent for the tested self-signed/cross-signed pair with unequal ranking), or define a stable tie-breaker if general order independence is intended.

### Existing discussion
I do not re-raise the cycle-prefix behavior or absence of complete path validation as new blockers: #339 explicitly accepted packaging a signature-verified nonrepeating prefix, and broader trust/constraint policy is tracked in #341. The concrete embedded-precedence regression above is separate from those agreed limitations.

Non-blocking test note: AddIgnoresDuplicateCertificates only asserts output count. It would still pass if Add retained duplicate certificates because issuer traversal selects one candidate per step; it does not establish internal deduplication. Also refresh the PR description to include validity ranking and 173 tests (six new cases).

### Verification
- Full original suite at 7f0fad9: **173 passed, 0 failed, 0 skipped** on macOS ARM64, SDK 10.0.301 / runtime 10.0.9.
- All four core-library targets built in Release.
- git diff --check passed; all five hosted checks are green.
- One additional temporary regression failed as described; removed afterward and the full original suite rerun. Working tree is clean.
- Known SourceLink submodule warnings remain. No local public-CA, Framework-runtime, or preferredChain HTTP end-to-end verification performed.

**AI-assisted review:** OpenCode using gpt-6-astra. Review of another session’s implementation with an independent diagnostic reproduction, not an independent human review.

fszlin:
## [P1] Embedded precedence — confirmed regression, fixed in `26b356c`

You are right, and this one is worse than a missed edge case: it contradicted a claim I made in the PR description. I wrote that "supplying only the cross-signed alternate still produces the long chain", and the merged ranking made that false whenever the self-signed variant is one the library embeds.

The cause is exactly as diagnosed. The old code consulted embedded certificates **only when the supplied lookup missed**:

```csharp
if (!certificates.TryGetValue(...) && !embeddedCertificates.Value.TryGetValue(...))
```

My `GetCandidates` yielded both sources into a single `OrderByDescending`, so ranking crossed the source boundary. A valid embedded self-signed root outranks a valid supplied cross-signed alternate on the self-signed tiebreak — silently changing embedded-root policy in a PR that claimed not to touch it.

Fixed by tiering: rank within supplied candidates, and fall back to embedded only when no supplied certificate can serve as issuer. Alternates and the validity/self-signed ranking still apply within each source.

Your point that **synthetic subject names cannot expose this** is the part I want to acknowledge specifically — it is why my own tests all passed. Every name in them is `CN=Certes …`, which can never collide with the embedded set, so the entire interaction was invisible to the suite. Two tests now use a subject and key the library actually embeds:

- `SuppliedIssuerTakesPrecedenceOverEmbeddedRoot` — builds a cross-signed variant carrying the embedded Fake LE Root X1 subject and public key, signed by an ephemeral root, supplies only that path, and asserts the embedded root is absent from the result. This is your reproduction, made permanent.
- `EmbeddedRootIsUsedWhenNoSuppliedIssuerServes` — pins the fallback that should still work, so the fix cannot be "disable embedded roots".

Merging the two sources back into one ranking fails the first and nothing else.

## [P2] Changelog claim — narrowed

Correct, and I overstated it. `TryGetIssuer` keeps insertion order for equal-ranked candidates, so two unexpired cross-signs under different roots still resolve by insertion order. The entry now states what actually holds — supplied-over-embedded precedence, then validity, then self-signed, with equal-ranked candidates resolving in the order added — rather than a blanket order-independence claim.

I did not invent a general tie-breaker. Ordering two equally valid cross-signs would need a policy input the store does not have, and `preferredChain` is the mechanism that already expresses it at the `Download` layer.

## Test note — agreed, it was vacuous

`AddIgnoresDuplicateCertificates` asserted only the output count, and you are right that it would pass without deduplication, since traversal selects one candidate per step. Deduplication has no observable effect on the chain at all, so rather than dress it up I renamed it to `AddingTheSameCertificateTwiceDoesNotAffectTheChain`, which is what it actually verifies. Deduplication remains an implementation detail.

## PR description

Refreshed: validity ranking, supplied-over-embedded precedence, 175 tests, 8 new cases, and the two additional mutations.

## Verification

- **175 passed, 0 failed, 0 skipped** on `net10.0`.
- Release build of all library targets; `net462` compiles; `git diff --check` clean.

| Mutation | Result |
| --- | --- |
| Restore last-write-wins in `Add` | 2 fail |
| Remove the self-signed preference | 1 fails |
| Remove the validity preference | 1 fails |
| Rank supplied and embedded candidates together | 1 fails |

Thanks for reproducing against real embedded material — that found a regression my synthetic fixtures were structurally incapable of catching.

fszlin (commented):
## Readiness re-review of 26b356c

**Ready to merge within the documented scope. Both findings from my previous review are addressed.**

- Supplied issuer selection now runs first; embedded selection is reached only via the null-coalescing fallback. An embedded self-signed root can no longer displace a supplied, signature-verifying cross-sign. The permanent regression reproduces my embedded-subject/key case and asserts the exact supplied path; a separate test preserves embedded fallback.
- The changelog now explicitly documents insertion-order resolution for equally ranked candidates, instead of claiming universal order independence.
- The PR description includes validity ordering, source precedence, and the updated test count. Renaming the duplicate-add test accurately describes its observable assertion.

No new blocking defect found. Full-path validation and embedded-root removal remain deferred to #341; the accepted partial-prefix behavior remains unchanged.

Non-blocking wording nit: the changelog sentence saying the self-signed variant is selected regardless of order should say “when validity ranking is equal.” Its preceding sentence correctly describes validity precedence, and the expired-self-signed regression correctly chooses the cross-sign instead. Export-level/preferredChain end-to-end coverage remains a useful follow-up; these new regressions exercise the shared store.

**Verification:** At 26b356c, independently re-ran the full net10.0 suite: **175 passed, 0 failed, 0 skipped**, macOS ARM64, SDK 10.0.301 / runtime 10.0.9. git diff --check passed. All five hosted checks passed; GitHub reports CLEAN merge status. Known SourceLink warnings remain. All-target Release builds were verified in my previous review and not repeated locally for this follow-up; public-CA, Framework-runtime and Azure App Service interoperability remain unverified. No source changes made.

**AI-assisted review:** OpenCode using gpt-6-astra; follow-up review of another session’s implementation, not an independent human review. This supersedes my prior changes-recommended verdict; no merge performed.

fszlin:
Approved. I accept the documented issuer-selection policy and the deferred scope tracked in #341. The review findings are resolved and all required checks are passing. Please merge.

PR #343, opened 2026-09-21
Stop shipping and consulting embedded CA roots

### LINKED ISSUE #341: Remove automatic embedded-root fallback and separate PFX packaging from path validation

fszlin:
Closes #341.

## Description

Two decisions, both confirmed before implementing.

### 1. Stop shipping and consulting embedded CA roots

`Resources/Certificates` carried three roots that were consulted automatically whenever a supplied issuer was missing:

| Resource | Status |
| --- | --- |
| `dst-root-ca-x3.pem` | **Expired 2021-09-30** |
| `fake-le-root-x1.pem` | **Let's Encrypt staging root** |
| `isrg-root-x1.pem` | Current |

Any of them could be appended to an exported chain without the caller asking — including an expired root, which is what broke OpenSSL 1.0.2 clients in 2021, and a staging trust anchor shipped in the production package. The set was Let's Encrypt-only and already stale, missing ISRG Root X2 and every other ACME CA, while Certes itself is CA-agnostic.

The fallback existed only to satisfy the walk-to-self-signed-root requirement, and #339 removed that requirement. Removal now degrades gracefully: the chain ends at the highest issuer available instead of failing.

This is the change [@aarongable asked for in #315](https://github.com/fszlin/certes/issues/315#issuecomment-5760729655).

### 2. `PfxBuilder` packages, it does not validate

`FindIssuers` ran `PkixCertPathBuilder` when a self-signed root happened to be present and packaged the linked issuers otherwise — so the same call validated or did not, depending on its inputs. It now always packages the certificate plus its linked issuers, excluding any self-signed root.

Issuers are still accepted only after verifying that they signed the certificate below them, which is what `CertificateStore` does. Issuer constraints, revocation and complete-path validity are the relying party's responsibility, and the XML docs now say so rather than implying otherwise.

## Breaking changes

1. **PEM export no longer appends a root the server did not return.** Callers who want one packaged supply it through the chain or `AddIssuer`/`AddIssuers`. For Let's Encrypt users this means `fullchain.pem` ends at the intermediate — which is what a TLS server should send anyway (RFC 8446 section 4.4.2).
2. **PFX no longer validates the path when a root is present.** Output shape is unchanged: the self-signed root was already excluded, pinned by `PfxChainExcludesRootWhenSupplied`.

Both are recorded under Breaking changes in the changelog. The next release already carries a breaking change for the CLI runtime, so these ride along rather than forcing a separate major.

## Verification

SDK 10.0.301, macOS ARM64.

- `dotnet test` — **177 passed, 0 failed, 0 skipped** (175 on `main`; 2 new).
- `dotnet build src/Certes/Certes.csproj -c Release` — all library targets. `net462` test project compiles; not run.
- `dotnet pack` and inspection of the packed assembly: **no certificate resources remain in the shipped DLL**.
- `git diff --check` clean.

Three tests changed, each documenting the intended behavior rather than being weakened:

- `CanGenerateFullChainPemWithKey` previously asserted that DST Root CA X3 was appended to the output. It now asserts no issuer is appended — the clearest statement of what changed.
- `EmbeddedRootsAreNotUsedAutomatically` replaces the embedded-precedence tests from #342, which are obsolete now that there is no embedded source. It uses `fake-le-intermediate-x1.pem`, whose issuer the library used to supply implicitly, and asserts the result is empty.
- `RootIsUsedWhenSuppliedExplicitly` pins that explicit roots still work, so the removal cannot be mistaken for dropping root support.

`ExportsExpiredCertificateWithoutValidatingThePath` pins the packaging-only policy itself: an expired certificate exports with and without a supplied root. This is the case the old implementation rejected, since PKIX checks validity, so it discriminates the policy rather than the output shape. Its assertion reads the PFX through .NET's `X509CertificateLoader` rather than BouncyCastle and signs with the imported key, so it does not share an implementation with the exporter.

| Mutation | Result |
| --- | --- |
| Remove the self-signed exclusion in `FindIssuers` | `PfxChainExcludesRootWhenSupplied` fails |
| Restore the PKIX branch when a root is present | `ExportsExpiredCertificateWithoutValidatingThePath(supplyRoot: True)` fails, and only that case |

**Not verified:** behavior against a real CA, `net462` at runtime, and whether any consumer depended on the embedded roots for a chain they could not otherwise assemble. That last one is the risk worth a second opinion — it is unobservable from this repository.

**Known gap, not introduced here:** export does not check that the supplied private key matches the certificate. Removing PKIX did not change this — the old path builder did not check it either — and the positive key-association assertions do not demonstrate rejection of mismatched inputs. Tracked in #344.

## Checklist

- [x] All tests are passing
- [x] Documentation updated: changelog breaking-change entries and `AGENTS.md` baseline

## AI assistance

**Review status:** Authored by this session, so the checks above are a self-review. No independent or human review is claimed. The scope of both removals was confirmed with the maintainer before implementation, but the user-impact question in "Not verified" deserves independent judgment.


fszlin (commented):
## Review of 07ca188

**No blocking implementation defect found in the reviewed diff. Not merge-ready yet: retarget to main and obtain required hosted checks.** The PR currently targets fix/cross-signed-alternates even though #342 has merged, and GitHub reports no checks. CLEAN against that feature branch is not verification of the merge into main.

The removal of embedded certificates is complete in the source/resource declarations. The simplified PFX path consistently packages signature-linked issuers and excludes the terminal self-issued/root certificate, matching the existing root-exclusion behavior. I accept the explicitly documented breaking policy of removing fallback for both PEM and PFX for this review; preserving PEM fallback was an earlier recommendation, not an implementation requirement after the reported maintainer scope decision.

### [P2] Add permanent coverage for the packaging-only behavior
`src/Certes/Pkcs/PfxBuilder.cs:111–121`, `test/Certes.Tests/Pkcs/PfxBuilderTests.cs`

No PfxBuilder test was added or changed to demonstrate the intentional removal of PKIX validation. Current valid-chain/root-exclusion cases also pass with the previous path builder. The reported root-filter mutation verifies exclusion, not the policy change. Add a signature-valid but expired leaf with a supplied root and assert export succeeds with the exact certificate and matching key, as it does without a root. This should fail if the old PKIX branch is restored. Include an independent .NET import/sign-verify assertion if practical; existing PFX assertions use the same BouncyCastle implementation as the exporter.

I tested this behavior temporarily and it works; this is regression coverage to retain, not a demonstrated production failure.

### Documentation corrections
- `docs/CHANGELOG.md` directs PEM users to PfxBuilder.AddIssuer/AddIssuers to package a root, but those APIs affect only PFX, which always excludes that root. Separate the guidance: add the root to CertificateChain.Issuers before ToPem for PEM output; AddIssuer supplies chain material to PfxBuilder but does not cause a self-signed root to be exported.
- The existing Unreleased cross-sign entry still says embedded certificates are consulted as fallback, contradicting the new breaking entry. Consolidate the unreleased notes to describe the final release behavior.
- The packaging-versus-validation explanation is on private FindIssuers. Put the contract on public Build/class/FullChain documentation too, and fix PfxChainExcludesRootWhenSupplied’s stale “used as a trust anchor” comment.
- Removing PKIX does not add private-key/certificate matching checks. That gap predates this PR; the positive key-association tests below do not demonstrate rejection of mismatched inputs. Track that separately rather than implying export validates all input relationships.

### Independent verification
- **175 original unit tests passed**, no failures/skips, on macOS ARM64, SDK 10.0.301 / runtime 10.0.9.
- **5 additional temporary review probes passed**: RSA and ES256 PFX imported using .NET X509CertificateLoader with and without a supplied root, exact two-certificate contents checked, imported private keys signed data successfully verified with the certificate public key; expired-leaf export/import worked both with and without a root.
- Initial probe runs used EphemeralKeySet, which macOS does not support; reran successfully with DefaultKeySet and disposed all imported certificates. This was a probe/platform issue, not a PR failure.
- Signed Release dotnet pack passed for all four library targets, using a fresh temporary NuGet cache and version 0.0.0-ci-review343; net10.0 PackageSmoke consumption passed (assembly identity, RSA/EC key round-trips, CSR).
- git diff --check passed. Temporary test source removed and original suite rerun; working tree clean.
- Known SourceLink and package-readme warnings remain. No local net8 runtime, Framework runtime, OpenSSL PFX import, public-CA or Azure App Service verification. The author’s packed-resource inspection was not independently repeated.

**AI-assisted review:** OpenCode using gpt-6-astra; review of another session’s implementation with independent interoperability probes, not an independent human review. No merge or base-branch modification performed.

fszlin:
All addressed in `bdab8d7`, and the PR is retargeted to `main`.

## Retarget and checks

#342 merged while this was open. Rebased onto `main` — note that a plain `git rebase origin/main` conflicts, because #342 was squash-merged and this branch still carried its original commits; `git rebase --onto origin/main 26b356c` replays only this PR's work cleanly. **Base is now `main` and all five required checks are green.**

Your point that "CLEAN against that feature branch is not verification of the merge into main" is worth underlining: the check rollup on the stacked PR was not merely unverified, it was *empty*, because `build.yml` triggers on `pull_request: branches: [main]` only. I had worked around that with `workflow_dispatch`, which produced five green jobs on the branch — but that ran the feature-branch code, not the merge result, so your objection stands regardless of that run.

## [P2] Permanent coverage for packaging-only behavior

Agreed, and the gap was exactly as described: every existing case passes under both implementations, so nothing pinned the *policy*.

`ExportsExpiredCertificateWithoutValidatingThePath` uses your suggested discriminator — a signature-valid but **expired** leaf, exported with and without a supplied root. PKIX checks validity, so the old branch rejected the `supplyRoot: true` case. Restoring that branch now fails that case and only that case:

```
MUTATION: restore the PKIX branch when a root is present
  Failed  ExportsExpiredCertificateWithoutValidatingThePath(supplyRoot: True)
  Failed!  - Failed: 1, Passed: 176
```

The `supplyRoot: false` case passes under both, which is what makes the pair a policy test rather than an output-shape test.

I took the independent-assertion suggestion too. The assertion reads the PFX through .NET's `X509CertificateLoader` rather than BouncyCastle, checks `RawData` against the expected leaf, and **signs with the imported private key and verifies against the certificate's public key** — so it does not share an implementation with the exporter. `DefaultKeySet` as you found; `EphemeralKeySet` is unsupported on macOS. Guarded with `#if NETFRAMEWORK` for the `net462` target, which has no `X509CertificateLoader`; that target compiles.

## Documentation corrections

- **PEM guidance was wrong.** `PfxBuilder.AddIssuer`/`AddIssuers` are PFX-only and never cause a self-signed root to be exported, so pointing PEM users at them was doubly misleading. Now: include the root in `CertificateChain.Issuers` before `ToPem`, with the PFX behavior stated separately.
- **Contradictory unreleased entries consolidated.** The cross-sign entry still described embedded certificates as a fallback. That sentence is removed, so the notes describe the final release behavior rather than the intermediate state of a merged PR.
- **Contract moved to the public surface.** The packaging-not-validation statement is now on the `PfxBuilder` class, `Build`, and `FullChain`, not only on the private helper. `PfxChainExcludesRootWhenSupplied`'s stale "used as a trust anchor" comment is corrected — that was left over from behavior this PR removes.
- **Key/certificate matching gap filed as #344**, not folded in. You are right that it predates this PR and that the positive assertions do not demonstrate rejection; the issue says so explicitly and notes the old PKIX path did not check it either. I have deliberately not implied that export validates input relationships.

## Verification

- **177 passed, 0 failed, 0 skipped** on `net10.0` (175 on `main`; 2 new).
- Release build of all library targets; `net462` test project compiles.
- `git diff --check` clean. Five green checks against `main`.

| Mutation | Result |
| --- | --- |
| Remove the self-signed exclusion in `FindIssuers` | `PfxChainExcludesRootWhenSupplied` fails |
| Restore the PKIX branch when a root is present | `ExportsExpiredCertificateWithoutValidatingThePath(supplyRoot: True)` fails, only that case |

Thanks for the independent import/sign probes and the packed-package consumption run — the expired-leaf discriminator in particular is a sharper test than anything I had written.

fszlin (commented):
## Readiness re-review of bdab8d7

**Ready to merge within the documented breaking-change scope. All items from my previous review are addressed.**

- The PR now targets main; all five required hosted checks are green and GitHub reports CLEAN merge status.
- Permanent expired-leaf tests cover both supplied-root and rootless export. They assert the exact packaged chain and independently load the PFX through .NET, checking the leaf and verifying a signature made by the imported private key. This protects the packaging-only policy rather than just normal valid-chain output.
- PEM migration guidance now correctly uses CertificateChain.Issuers, the contradictory embedded-fallback wording is removed, and the public PfxBuilder/Build/FullChain docs state the packaging contract.
- The pre-existing private-key mismatch validation gap is explicitly documented and tracked in #344 rather than claimed fixed.

I reviewed the revised diff and response comment, and independently reran the complete net10.0 suite at bdab8d7: **177 passed, 0 failed, 0 skipped** on macOS ARM64 (SDK 10.0.301 / runtime 10.0.9). git diff --check passed. Known SourceLink warnings remain. Signed Release packing, fresh-cache package consumption, and additional RSA/EC .NET import probes passed in my previous review; those were not repeated for this documentation/test follow-up. Framework runtime, OpenSSL import, public-CA, and Azure App Service interoperability remain unverified locally.

No remaining blocking findings. No source changes or merge performed.

**AI-assisted review:** OpenCode using gpt-6-astra; follow-up review of another session’s implementation, not an independent human review. This supersedes the previous not-merge-ready assessment.

